PR #9, opened 2023-10-12
Initial MCH simulation support

hintxiv:
Untested but hopefully working :)

* Uses an infinite duration damage model that scales with GCD recast
* Auto damage is slightly inaccurately modelled, requires a separate math path
* WF damage is slightly inaccurately modelled, requires a separate math path (DoTs)


xpdota:
You can also add it to getDefaultSims() if you want it to be added automatically to level 90 MCH sheets.

PR #10, opened 2023-10-16
Initial AST sim implementation

Mayceia:
Created an AST sim based on Wynn's SGE sim MK2
Not complete, doesn't use the extra astrodyne every 6 minutes


xpdota:
Thanks!

PR #13, opened 2023-10-23
Buff logic rework

Mayceia:
Changes how buffs are timed, so each is now on its own timer, and reworked the healer cycle processor sims for the new buff logic.


xpdota:
Succeeded by https://github.com/xiv-gear-planner/gear-planner/pull/14

PR #20, opened 2024-01-03
WIP: Autoattack support

xpdota:
Cycle sims will have auto-attacks automatically factored in.

### DIFF
--- a/src/scripts/xivmath.ts
+++ b/src/scripts/xivmath.ts
@@ -11,6 +11,7 @@ import {AttackType, ComputedSetStats, JobData, LevelStats} from "./geartypes";
     object. It is preferable to use already-computed values since those are also exposed on the UI. This ensures
     consistency with the displayed values.
  */
+const fl = Math.floor;
 
 /**
  * Convert skill speed to GCD speed.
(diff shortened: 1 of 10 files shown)


src/scripts/xivmath.ts
```diff
@@ -180,19 +181,43 @@ export function mpTick(levelStats: LevelStats, piety: number) {
...
  *
  * TODO: where to factor in extra damage from buffs?
  */
-export function baseDamage(stats: ComputedSetStats, potency: number, attackType: AttackType = 'Unknown', autoDH: boolean = false, autoCrit: boolean = false) {
+export function baseDamage(stats: ComputedSetStats, potency: number, attackType: AttackType = 'Unknown', autoDH: boolean = false, autoCrit: boolean = false, isDot: boolean = false) {
 
+    let spdMulti: number;
+    const isAA = attackType === 'Auto-attack';
+    if (isAA) {
+        spdMulti = stats.sksDotMulti;
```
furstblumier, line 208: afaik (but not really tested, not that it makes a real diff) it's sps for casters here

xpdota: image

I think it's sks for both, unless the tooltip is wrong

PR #31, opened 2024-01-27
add hp calculation

furstblumier:


src/scripts/xivconstants.ts
```diff
@@ -431,8 +441,8 @@ export const LEVEL_ITEMS: Record<SupportedLevel, LevelItemInfo> = {
 /**
  * Main stats in current version of the game.
  */
-export const MAIN_STATS = ['strength', 'dexterity', 'intelligence', 'mind', 'vitality'] as const;
-// TODO: is Tenacity treated like this? No.
+export const MAIN_STATS = ['strength', 'dexterity', 'intelligence', 'mind', 'vitality', 'hp'] as const;
```
xpdota, line 444: Why does it have to be added here?

PR #37, opened 2024-02-04
Party bonus apply to vit

xpdota:


furstblumier:
if `combinedStats` contains the base vit it lgtm

PR #46, opened 2024-04-07
Additional job traits

xiashtra:


xiashtra:
Included as part of #48.

Issue #43, opened 2024-04-07
Ability cast time is not appropriately reduced by sps/sks/haste


xiashtra:
Currently, the sim adjusts ability gcd time by sps/sks/haste as appropriate, but does not adjust ability cast time in the same way.  For example, a 2.5s gcd with a 2.00s cast time reduced by spell speed to a 2.20s gcd should have a cast time of 1.76s, but the sim instead reduces it to a 2.20s gcd with a 2.00s cast time.

xiashtra:
Closing as addressed by #47.

PR #47, opened 2024-04-07
Bugfixes for cast times and application delay

### LINKED ISSUE #42: [bug] Abilities apply buffs too early
Currently, abilities that apply a buff activate that buff at the ability snapshot time, instead of the end of the ability cast.  This causes buffs to apply too early, and for abilities to snapshot their own buffs before they are finished casting.  This can be seen in the current RPR sim, where the initial use of Shadow of Death is benefitting from the Death's Design buff before the buff should be active.

ref: https://github.com/xiv-gear-planner/gear-planner/blob/530499cde768b96be91366f23ac8dd45dab30f46/src/scripts/sims/sim_processors.ts#L459-L462

https://github.com/xiv-gear-planner/gear-planner/blob/530499cde768b96be91366f23ac8dd45dab30f46/src/scripts/sims/sim_processors.ts#L502-L504

xpdota:
Fixes #42 and #43 

Additionally adds hover text to the time column in the abilities used table.

### DIFF
--- a/src/scripts/sims/common/swiftcast.ts
+++ b/src/scripts/sims/common/swiftcast.ts
@@ -1,0 +1,29 @@
+import {Ability, Buff, BuffController, OgcdAbility} from "../sim_types";
+
+export const SwiftcastBuff: Buff = {
+    cooldown: 0,
+    duration: 0,
+    effects: {},
+    // TODO
+    job: 'WHM',
+    name: "Swiftcast",
+    selfOnly: true,
+    startTime: 0,
+    beforeAbility: (controller: BuffController, ability: Ability) => {
+        if (ability.type === 'gcd' && ability.cast >= 0) {
+            ability.cast = 0;
+            controller.removeStatus(SwiftcastBuff);
+        }
+    }
+}
+
+export const SwiftcastAbility: OgcdAbility = {
+    activatesBuffs: [SwiftcastBuff],
+    id: 0, // TODO
+    name: "Swiftcast",
+    potency: null,
+    type: "ogcd",
+    attackType: 'Ability',
+    animationLock: 0.6
+
+}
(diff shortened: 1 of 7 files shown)


xpdota:
@xiashtra I have deployed this on [https://beta.xivgear.app/](https://beta.xivgear.app/), could you please take a look and see if it fixes the issues?

src/scripts/sims/common/swiftcast.ts
```diff
@@ -0,0 +1,29 @@
...
+        if (ability.type === 'gcd' && ability.cast >= 0) {
+            ability.cast = 0;
+            controller.removeStatus(SwiftcastBuff);
+        }
+    }
+}
+
+export const SwiftcastAbility: OgcdAbility = {
+    activatesBuffs: [SwiftcastBuff],
+    id: 0, // TODO
```
xiashtra, line 22:
```suggestion
    id: 7561,
```

src/scripts/sims/common/swiftcast.ts
```diff
@@ -0,0 +1,29 @@
+import {Ability, Buff, BuffController, OgcdAbility} from "../sim_types";
+
+export const SwiftcastBuff: Buff = {
+    cooldown: 0,
+    duration: 0,
```
xiashtra, line 5:
```suggestion
    cooldown: 60,
    duration: 10,
```

xpdota: This isn't actually referenced anywhere, still a WIP, but thank you!

xiashtra:
I believe this function will need to take into account gcd reduction also: https://github.com/xiv-gear-planner/gear-planner/blob/2f3d19643fd10473fe5aa2d08463df373ed7701c/src/scripts/sims/sim_processors.ts#L34-L40

xpdota:
Ok, I redid application delay. The appDelay function now returns the delay from snapshot to application, rather than start-of-cast to application. It really only still exists as a function so that I can later add an application delay field to abilities to let them individually override.

PR #48, opened 2024-04-10
Implement BLU weapon damage modifier

xiashtra:
Expects #46.

Implements the BLU weapon damage modifier as described in ["BLU Mage Damage Formula"](https://docs.google.com/document/d/1-NBJDkyl8h_UXFcOba-CrPPcFXZrDs2QXX53pSS3QJc) and discussions from the Allagan Studies Discord.  Also corrects a bug that was filtering out all BLU weapons from gear selection.

Known general limitations:
- The current Xivgear damage formula implementation uses non-caster order of operations.  This causes small amounts of error for all caster actions (limited testing shows <1% deviation).
- The current Xivgear damage formula implementation does not account for the low potency (<100) modifier.  This causes small amounts of error for all jobs with <100 potency actions (limited testing shows <1% deviation).

### DIFF
--- a/src/scripts/datamanager.ts
+++ b/src/scripts/datamanager.ts
@@ -128,7 +128,7 @@ export class DataManager {
                 // Basic item properties
                 'ID', 'IconHD', 'Name', 'LevelItem',
                 // Equip slot restrictions
-                'EquipSlotCategory', 'IsUnique',
+                'ClassJobCategory', 'EquipSlotCategory', 'IsUnique',
                 // Stats
                 'Stats', 'DamageMag', 'DamagePhys', 'DelayMs',
                 // Materia
@@ -166,8 +166,10 @@ export class DataManager {
                 }
             }).then((rawItems) => {
                 this.allItems = rawItems
-                    // TODO: filtering out items with no stats excludes all BLU weapons
-                    // .filter(i => i['Stats'] !== null)
+                    .filter(i => {
+                        return i['Stats'] !== null
+                            || (i['ClassJobCategory']?.['BLU'] === 1 && i['EquipSlotCategory']?.['MainHand'] === 1); // Don't filter out BLU weapons
+                    })
                     .map(i => new XivApiGearInfo(i));
                 // TODO: put up better error
             }, (e) => console.error(e));
--- a/src/scripts/datamanager.ts
+++ b/src/scripts/datamanager.ts
@@ -166,7 +166,8 @@ export class DataManager {
                 }
             }).then((rawItems) => {
                 this.allItems = rawItems
-                    .filter(i => i['Stats'] !== null)
+                    // TODO: filtering out items with no stats excludes all BLU weapons
+                    // .filter(i => i['Stats'] !== null)
                     .map(i => new XivApiGearInfo(i));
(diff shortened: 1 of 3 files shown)


src/scripts/xivconstants.ts
```diff
@@ -628,4 +634,42 @@ export function formatAcquisitionSource(source: GearAcquisitionSource): string |
...
+    [550, 77], [570, 78], [590, 79], [620, 80], [650, 81], [680, 82], [710, 83],
+    [740, 84], [750, 85], [770, 86], [790, 87], [810, 88], [830, 89], [860, 90],
+    [880, 91], [900, 92], [930, 93], [960, 94], [990, 95], [1030, 96], [1060, 97],
+    [1080, 98], [1110, 99], [1140, 100], [1170, 101], [1200, 102], [1230, 103],
+    [1260, 104], [1290, 105],
+    // TODO: the following are predicted values for lvl90 BLU, will need to be verified
+    [1340, 106], [1360, 107], [1390, 111], [1510, 113], [1590, 115], [1680, 117],
+    [1780, 119], [1880, 121], [1980, 123], [2090, 125], [2200, 127], [2320, 129],
+    [2410, 131]
+] as const as readonly number[][];
```
xpdota, line 660:
```suggestion
] as const as readonly (readonly number[])[];
```

Come to think of it, does this need to be exported at all? Isn't it only accessed by `bluWdfromInt`?

src/scripts/datamanager.ts
```diff
@@ -166,7 +166,8 @@ export class DataManager {
                 }
             }).then((rawItems) => {
                 this.allItems = rawItems
-                    .filter(i => i['Stats'] !== null)
+                    // TODO: filtering out items with no stats excludes all BLU weapons
+                    // .filter(i => i['Stats'] !== null)
```
xpdota, line 170:
```suggestion
                    .filter(i => {
                        return i['Stats'] !== null
                            || (i['ClassJobCategory']?.['BLU'] === 1 && i['EquipSlotCategory']?.['MainHand'] === 1);
                    })

```

GH won't let me leave a suggestion for something that's not actually part of the PR, but you'd have to add `'ClassJobCategory'` to the columns that it fetches.

src/scripts/xivconstants.ts
xiashtra: You're right, it doesn't, I think I had that there from earlier before I put the function into `xivconstants.ts` also.

src/scripts/datamanager.ts
xiashtra: Would that be the preferred approach?  Right now, the hard limits on item levels set in `xivconstants.ts` is preventing any of the glamour gear from getting fetched, so the filter is really only excluding the BLU weapons as far as I can see.

xpdota: I think it's the safest option, since it limits the scope of the change as much as possible.

xiashtra: Ok.  I was just thinking in terms of data size of the return request, adding `ClassJobCategory` to every item is going to have a much bigger impact.

xpdota:
Thank you!

PR #50, opened 2024-04-12
initial cooldown manager support

### LINKED ISSUE #45: [feature request] Ability cooldown tracking
It would be useful for job modeling if the sim supported tracking cooldowns of abilities, to ensure they are not used before the cooldown timer has expired.  This would include both fixed cooldowns, and cooldowns that can be reduced by sps/sks/haste.

xpdota:
Adds cooldown manager and initial tests

Fixes #45


src/scripts/sims/sim_processors.ts
```diff
@@ -266,8 +267,13 @@ export class CycleProcessor {
     private readonly manuallyActivatedBuffs: readonly Buff[];
     combatStarted: boolean = false;
     readonly useAutos: boolean;
+    private readonly cdTracker: CooldownTracker;
```
xiashtra, line 270:
```suggestion
    readonly cdTracker: CooldownTracker;
```
If this is `private`, it can't be queried from a sim to check the cooldown status of an ability before trying to use it.

xpdota:
@xiashtra anything else before I merge?

xiashtra:
Seems to be working ok for fixed cooldowns, just need support for the ones reduced by sps/sks.

PR #61, opened 2024-04-17
Initial BLU Winged 120s implementation

xiashtra:
Expects #58.

Implements a simulation for a BLU Winged Reprobation rotation with burst windows every 120s.
Portions of the sim are abstracted with the intent to share components between multiple BLU sims (to be submitted as separate PRs).

### DIFF
--- a/src/scripts/sims/blu_common.ts
+++ b/src/scripts/sims/blu_common.ts
@@ -576,13 +576,13 @@ export const BeingMortal: OgcdAbility = {
     id: 34582
 }
 
-const WaningInfo: OgcdAbility = { // dummy ability to insert info into timeline
-    name: "-- Waning --",
-    type: "ogcd",
-    attackType: "Ability",
-    potency: null,
-    animationLock: 0
-}
+// const WaningInfo: OgcdAbility = { // dummy ability to insert info into timeline
+//     name: "-- Waning --",
+//     type: "ogcd",
+//     attackType: "Ability",
+//     potency: null,
+//     animationLock: 0
+// }
 
 /**
  * BLU sim settings
@@ -702,16 +702,17 @@ export abstract class BluSim<_BluCycleSimResult, _BluSimSettings, BluExternalCyc
     }
 
     doWaning(cp: CycleProcessor) {
-        const waxingEnd = this.rotationState.fluteStart + WaxingNocturne.duration;
+        // const waxingEnd = this.rotationState.fluteStart + WaxingNocturne.duration;
         const waningEnd = this.rotationState.fluteStart + WaxingNocturne.duration + WaningNocturne.duration;
+        cp.addSpecialRow("-- Waning End --", waningEnd);
 
-        // advance to end of Waxing window
-        cp.advanceTo(Math.max(cp.currentTime, Math.min(waxingEnd, cp.totalTime)), true);
+        // // advance to end of Waxing window
+        // cp.advanceTo(Math.max(cp.currentTime, Math.min(waxingEnd, cp.totalTime)), true);
 
-        // insert informational entry into timeline
-        if (waningEnd > cp.currentTime) {
-            cp.use(WaningInfo);
-        }
+        // // insert informational entry into timeline
+        // if (waningEnd > cp.currentTime) {
+        //     cp.use(WaningInfo);
+        // }
 
         // advance to end of Waning window
         cp.advanceTo(Math.max(cp.currentTime, Math.min(waningEnd, cp.totalTime)), true);
@@ -742,12 +743,14 @@ export abstract class BluSim<_BluCycleSimResult, _BluSimSettings, BluExternalCyc
             case "Phantom Flurry": { // 5s channel time
                 cp.use(ability);
                 const endFlurry = cp.currentTime + 5;
+                cp.addSpecialRow("-- Flurry End --", endFlurry);
                 cp.advanceTo(Math.max(cp.currentTime, Math.min(endFlurry, cp.totalTime)), true);
                 break;
             }
             case "Apokalypsis": { // 10s channel time
                 cp.use(ability);
                 const endApok = cp.currentTime + 10;
+                cp.addSpecialRow("-- Apokalypsis End --", endApok);
                 cp.advanceTo(Math.max(cp.currentTime, Math.min(endApok, cp.totalTime)), true);
                 break;
             }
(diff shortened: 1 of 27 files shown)


src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+}
+
+const WaningNocturne: Buff = {
+    name: "Waning Nocturne",
+    job: "BLU",
+    duration: 15,
+    cooldown: 0,
+    selfOnly: true,
+    effects: {
+        dmgIncrease: -1 // can't use any actions during Waning
```
xpdota, line 108: Would it help if I added an 'actions blocked' property and a 'wait until actions unblocked' method?

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+    duration: 30,
+    cooldown: 0,
+    selfOnly: true,
+    effects: {
+        dmgIncrease: 0.8 // only applies to physical damage spells
+    },
+    appliesTo: ability => ability.attackType === "Spell" && BLU_PHYSICAL_SPELLS.includes(ability.name),
+    beforeSnapshot<X extends Ability>(buffController: BuffController, ability: X): X {
+        // mutually exclusive with Boost
+        buffController.removeStatus(Boost);
```
xpdota, line 125: I could add a 'mutually exclusive buffs' property

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+    selfOnly: true,
+    effects: {
+       // increases base potency of physical damage spells by 100 per hit
+    },
+    beforeSnapshot<X extends Ability>(buffController: BuffController, ability: X): X {
+        if (ability.attackType === "Spell" && BLU_PHYSICAL_SPELLS.includes(ability.name)) {
+            buffController.removeSelf();
+
+            // Triple Trident hits 3 times, each hit gets bonus potency
+            const bonusPotency: number = (ability.name === "Triple Trident") ? 300 : 100;
```
xpdota, line 206: I could add a 'number of hits' property to abilities

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+    name: "Phantom Flurry",
+    type: "ogcd",
+    attackType: "Ability",
+    potency: null,
+    dot: {
+        tickPotency: 200,
+        duration: 18, // TODO: ticks once on use, and 1 tick per second afterwards for 6 total ticks over 5s
+        id: 2502
+    },
+    animationLock: 0, // TODO: 5s channeled attack, prevents autos for duration
```
xpdota, line 470: I wonder what the best way to handle flurry would be. Especially since it would be extremely sensitive to catching or missing ticks.

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+        reducedBy: "spellspeed"
+    },
+    id: 34576
+}
+
+export const MortalFlame: GcdAbility = {
+    name: "Mortal Flame",
+    type: "gcd",
+    attackType: "Spell",
+    potency: null,
```
xpdota, line 527: This should be 0 - null makes it NonDamagingAbility which doesn't have a dot field.

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+    attackType: "Ability",
+    potency: 800,
+    animationLock: 0.6,
+    cooldown: {
+        time: 120 // shared with Apokalypsis
+    },
+    id: 34582
+}
+
+const WaningInfo: OgcdAbility = { // dummy ability to insert info into timeline
```
xpdota, line 579: There's an existing `SpecialRecord` type for this which is used for the 'start of cycle' messages. I could add a method to let sims add arbitrary messages.

src/scripts/sims/blu_common.ts
xiashtra: Under the hood, Flurry appears to be a code copy of MCH Flamethrower, in that it uses the non-caster dot OoO for damage calculation.  I implemented it as a fake dot so it would pick up the spell-speed bonus on ticks appropriately, which was more important than having the tick timing be exactly modeled.

src/scripts/sims/blu_common.ts
```diff
@@ -0,0 +1,820 @@
...
+        }
+
+        if (this.settings.basicInstinctEnabled) {
+            cp.activateBuff(BasicInstinct);
+        }
+    }
+
+    useFlute(cp: CycleProcessor) {
+        cp.use(MoonFlute);
+        this.rotationState.fluteStart = cp.currentTime;
```
xpdota, line 699: I feel like it would be cleaner if I added a `makeCycleProcessor` method, so that you can supply a subclass of CycleProcessor that handles this behavior internally, rather than needing to manually call it.

src/scripts/sims/blu_common.ts
xiashtra: I don't think any other job would have a use for it currently, and the method I have to handling Waning seems to work ok right now, so it's not a priority.

src/scripts/sims/blu_common.ts
xiashtra: That would be useful for this case.

src/scripts/sims/blu_common.ts
xiashtra: If other jobs get segmented attacks in Dawntrail that might be worth having, but I wouldn't consider it a priority at the moment.

src/scripts/sims/blu_common.ts
xiashtra: That might be nice for adding some informational messages into the rotation.

xpdota: https://github.com/xiv-gear-planner/gear-planner/pull/62

src/scripts/sims/blu_common.ts
xpdota: https://github.com/xiv-gear-planner/gear-planner/pull/63

xpdota: I realized this is actually necessary rather than just convenient. The Sim instance is re-used for multiple runs of the same simulation, so anything that needs to be scoped to a specific run should be moved to the CycleProcessor or some other object that only lives within the simulate() method.

xiashtra:
Merged in #62 and added some `Special Record` uses.

src/scripts/sims/blu_common.ts
xiashtra: Merged #63 and refactored rotation state into custom `CycleProcessor`

xpdota:
Thanks so much!

PR #62, opened 2024-04-17
Special record support

xpdota:
Allow manual insertion of SpecialRecord

Known issue is that if you manually specify a time, it won't re-order the resulting list to accomodate.


xpdota:
@xiashtra

PR #63, opened 2024-04-17
Custom cycle processor support

xpdota:
- Adds support for custom subclasses and custom creation of CycleProcessor
- This acts as a convenient place to hold 'rotation state' rather than doing it on the Sim class directly
- Split up sim_tests.ts

### DIFF
--- a/src/scripts/test/sims/sim_tests.ts
+++ b/src/scripts/test/sims/sim_tests.ts
@@ -267,7 +267,7 @@ describe('Cycle sim processor', () => {
 });
 
 class CustomCycleProcessor extends CycleProcessor {
-    fillerCount: number = 0;
+    private fillerCount: number = 0;
 
     use(ability: Ability): AbilityUseResult {
         const out = super.use(ability);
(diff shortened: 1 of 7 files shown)


src/scripts/test/sims/sim_tests.ts
```diff
@@ -0,0 +1,338 @@
...
+                outer.fillerCount = cp.count;
+            }
+        }];
+    }
+}
+
+export const testCustomSimSpec: SimSpec<TestCustomMultiCycleSim, TestSimSettingsExternal> = {
+    displayName: "Test Custom Sim",
+    loadSavedSimInstance(exported: TestSimSettingsExternal) {
+        return new TestMultiCycleSim(exported);
```
xiashtra, line 320:
```suggestion
        return new TestCustomMultiCycleSim(exported);
```

xiashtra:
Maybe `settings` shouldn't be private on `CycleProcessor` so they can be read from custom CycleProcessors also.

PR #65, opened 2024-04-19
Fix bug in auto-attack damage calculation

xiashtra:
`f(AUTO)` should use the appropriate job modifier for auto attacks.


xpdota:
Thanks!

PR #67, opened 2024-04-19
Split buffs into personal and party types

### LINKED ISSUE #60: Refactor: Split party buffs and personal buffs
Currently, there is a single 'Buff' type, which is used for both personal buffs applied by abilities, such as Swiftcast, as well as party buffs which are applied on a schedule (if not simming its class), such as Chain.

Instead, 'Buff' should be the bare minimum for a status effect, while 'PartyBuff' could have the extra fields, such as cooldown and job.

xpdota:
Splits Buffs into BaseBuff and PartyBuff so that personal-only buffs do not need to specify useless fields such as job and CD.

Closes #60

### DIFF
--- a/src/scripts/sims/sim_types.ts
+++ b/src/scripts/sims/sim_types.ts
@@ -207,26 +207,15 @@ export type BuffController = {
 }
 
 // TODO: refactor this into Buff and PartyBuff so that the fields used for the party buffs (job, CD) can be untangled
-export type Buff = Readonly<{
+export type BaseBuff = Readonly<{
     /** Name of buff */
     name: string,
-    /** Job of buff */
-    job: JobName,
-    /** "Optional" would be things like DNC partner buffs, where merely having the job
-     // in your comp does not mean you would necessarily get the buff. */
-    optional?: boolean,
     /** Can only apply to self - not a party/targeted buff */
     selfOnly?: boolean,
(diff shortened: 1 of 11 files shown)


src/scripts/sims/sim_types.ts
```diff
@@ -207,26 +207,15 @@ export type BuffController = {
...
+export type BaseBuff = Readonly<{
     /** Name of buff */
     name: string,
-    /** Job of buff */
-    job: JobName,
-    /** "Optional" would be things like DNC partner buffs, where merely having the job
-     // in your comp does not mean you would necessarily get the buff. */
-    optional?: boolean,
     /** Can only apply to self - not a party/targeted buff */
     selfOnly?: boolean,
```
xiashtra, line 214: `selfOnly` seems redundant on `BaseBuff` now if `BaseBuff` will only be used for personal buffs.

xiashtra: Though thinking about it more, I guess you would still have buffs like Chain used manually in the SCH sim, which aren't `selfOnly`.

xpdota: Yeah, that's the intention. `manuallyActivatedBuffs` is the mechanism for turning party buffs into self-buffs.

PR #70, opened 2024-04-20
Settings for how to trim cycle duration to re-align

xpdota:


xiashtra:
Initial test of `full-duration` works as expected.

PR #71, opened 2024-04-21
Add support for casted oGCDs

### LINKED ISSUE #56: Feature: oGCDs with cast times
BLU has a few oGCDs with cast times which need to be accounted for.

xpdota:
Unifies useGcd and useOgcd, moves cast-related fields up to BaseAbility.

This allows oGCDs to have a cast time.

Closes #56


xiashtra:
This introduced a bug for oGCDs with `animationLock` set to a value different than `STANDARD_ANIMATION_LOCK` no longer being respected.

PR #73, opened 2024-04-22
Stats server

xpdota:
Very rough implementation of a server component that calculates set stats, possibly sims and such later.

The project is also broken into separate workspace. Currently, there is one for xivmath, which includes many of the basic gear types, one for the frontend code, and one for the stats server. The intention is to eventually break out parts into a fourth module, for frontend-specific things. That was, the server component won't have to depend on the frontend module at all.

This includes changes from #72


xiashtra:
Successfully built and ran this branch in VSCode, no obvious issues.  `npm serve` and `npm test` scripts also ran without issue.

PR #78, opened 2024-04-23
BLU ilvl display fix

xpdota:
Level 80's default item display settings are tailored for TEA since that's the most common reason you would use a level 80 sync. BLU relevant gear is 520-535.


xiashtra:
Looks good.

PR #85, opened 2024-04-25
Initial BLU Breath 60s implementation

xiashtra:
Expects #84.

Implements a simulation for a BLU Breath of Magic rotation with burst windows every 60s.


src/scripts/sims/blu_winged120.ts
```diff
@@ -179,7 +190,16 @@ export class BluWinged120Sim extends blu.BluSim<BluWinged120SimResult, BluWinged
                         cycle.use(blu.TripleTrident);
                     } else {
                         // otherwise, finish off the fight with a Final Sting combo
-                        cycle.use(blu.FeculentFlood);
+                        if (cp.cdTracker.canUse(blu.RoseOfDestruction, cp.nextGcdTime)) {
```
xpdota, line 193: I think I will add a method to the CycleProcessor to handle this so that you don't have to explicitly check at the GCD time.

PR #92, opened 2024-04-29
Add simulation description field and UI

xpdota:
image


xiashtra:
For some reason, this doesn't show correctly if I use the Live Server extension to test, but if I run the `serve` script, it does.

xpdota:
> For some reason, this doesn't show correctly if I use the Live Server extension to test, but if I run the `serve` script, it does.

Hmm, wonder if it's a stale cache or something

PR #98, opened 2024-05-09
Simulate stddev

xpdota:
The results display now shows the normal EV, as well as +1, +2, and +3:

image

The cycle sim config UI now has an additional field at the bottom to choose which one to pick as the result that gets shown on the table:

image

This accepts non-integer as well as negative values.


packages/xivmath/src/xivmath.ts
```diff
@@ -330,8 +347,38 @@ export function applyDhCrit(baseDamage: number, stats: ComputedSetStats) {
...
 
+export function dhCritPercentStdDev(stats: ComputedSetStats) {
+    // Use 1 as a base so that this acts as a multiplier
+    const baseDamage = 1.0;
+    const critDamage = baseDamage * stats.critMulti;
+    const dhDamage = baseDamage * stats.dhitMulti;
+    // Crit+DH multiplier
+    const cdhDamage = baseDamage * stats.critMulti * stats.dhitMulti;
+    // Crit+DH chance
+    const cdhChance = stats.critChance * stats.critMulti;
```
xiashtra, line 358:
```suggestion
    const cdhChance = stats.critChance * stats.dhitChance;
```

xiashtra (changes requested):

packages/xivmath/src/xivmath.ts
xpdota: Thanks

PR #101, opened 2024-05-14
Property to allow autoattacks to be disabled by default in a cycle sim

### LINKED ISSUE #100: Feature: Ability to toggle auto-attacks off by default for sim
Request a way to toggle auto-attacks off by default for a sim implementation.  Several jobs (casters, healers) do not have auto-attacks enabled by default, and/or they have abilities that specifically disable auto-attacks when used to prevent them from interfering with job mechanics (BLU).  While it is possible to force autos on those jobs, this is usually done through third-party tools and does not represent how the majority of players will interact with those jobs.

xpdota:
Closes #100 

Override `get useAutosByDefault()` to control behavior.

In a future version, it could automatically disable for all healers and casters.


xpdota:
@xiashtra

PR #109, opened 2024-05-23
Decouple UI and calc logic, part 1

### LINKED ISSUE #59: Long-Term Refactor: Untangle calculations from UI code
Currently, the API is just a k/v store, where only the user-selected bits of data (e.g. items, materia, sim settings) are stored and retrieved.

Some API consumers would like to also have computed values, such as total set stats and possibly simulation results. If the code were split into UI and non-UI code, then the non-UI portion could be adapted into a microservice that provides computed results for consumers who need it.

xpdota:
Nearly completely untangles the UI from backend code.

Most things have moved around a bit. 

There is now a fourth module, "core", which is intended for things that are not frontend but also not directly tied enough to in-game math to go in xivmath.

Closes #59 (mostly)

There's still a *small* dependency between core and DOM (due to simulations returning their configuration as an HTMLElement), but the memory consumption is still heavily reduced (see comment).


xpdota:
Memory reduction for stats server:
```
# Old
xivgear         xivgear-xivgear-stats-api-5fb7b8fbd5-kmt5b       1m           384Mi
xivgear         xivgear-xivgear-stats-api-5fb7b8fbd5-x6p2q       1m           375Mi
# New
xivgear         xivgear-xivgear-stats-api-beta-ff6c76c8d-mzctj   1m           124Mi
```

xpdota:
Currently on https://btest.xivgear.app/

PR #126, opened 2024-06-09
DNCsim

xpdota:
- Implements BaseUsageCountSim, a sim that operates by having the sim implementor provide the # of each skill that fits in each length of buff window
- Implements DNC sim using this new base sim

### DIFF
--- a/packages/xivmath/src/xivconstants.ts
+++ b/packages/xivmath/src/xivconstants.ts
@@ -213,6 +213,18 @@ export const JOB_DATA: Record<JobName, JobDataConst> = {
     MCH: STANDARD_RANGED,
     DNC: {
         ...STANDARD_RANGED,
+        traitMulti: (level, attackType) => {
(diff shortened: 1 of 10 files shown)


packages/xivmath/src/xivconstants.ts
```diff
@@ -213,6 +213,18 @@ export const JOB_DATA: Record<JobName, JobDataConst> = {
...
+                return 1;
+            }
+            else if (level >= 60) {
+                return 1.2;
+            }
+            else if (level >= 50) {
+                return 1.2;
+            }
+            return 1;
+        },
```
xiashtra, line 227: This trait modifier is already handled in line 145 above.  All Physical Ranged have the same trait.

xiashtra (changes requested):

packages/xivmath/src/xivconstants.ts
xpdota: Removed

PR #133, opened 2024-06-15
Pot/tincture buffs

xpdota:
See `potion.ts` for an example. I've only done grade 8 mind tinc so far for testing.

This PR also makes two wider changes:

- Pre-pull buffs are now timed correctly
- ID is now a required field for abilities. All existing abilities have had their IDs added.

### DIFF
--- a/packages/core/src/sims/common/potion.ts
+++ b/packages/core/src/sims/common/potion.ts
@@ -1,0 +1,46 @@
+import {OgcdAbility} from "../sim_types";
+import {fl, mainStatMulti} from "@xivgear/xivmath/xivmath";
+import {RawStatKey} from "@xivgear/xivmath/geartypes";
+import {capitalizeFirstLetter} from "../../util/strutils";
+
+function applyPotionBuff(initialValue: number, bonus: number, cap: number): number {
+    const effectiveBonus = Math.min(fl(initialValue * bonus), cap);
+    return initialValue + effectiveBonus;
+}
+
+function makeTincture(grade: number, stat: RawStatKey, id: number, itemId: number, bonus: number, cap: number): Readonly<OgcdAbility> {
+    const statName = capitalizeFirstLetter(stat);
+    return {
+        name: `Grade ${grade} Tincture of ${statName}`,
+        // TODO: ID is an item, not an ability - can possibily use attackType === item to check
+        // if the icon needs to be an item ability?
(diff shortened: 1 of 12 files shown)


packages/core/src/sims/common/potion.ts
```diff
@@ -0,0 +1,47 @@
...
+                name: 'Medicated',
+                duration: 30,
+                statusId: 0x31,
+                effects: {
+                    modifyStats: original => {
+                        const stats = {...original};
+                        stats[stat] = applyPotionBuff(stats[stat], bonus, cap);
+                        if (stat === stats.jobStats.mainStat) {
+                            stats.mainStatMulti = mainStatMulti(stats.levelStats, stats.jobStats, stats[stats.jobStats.mainStat]);
+                            stats.aaStatMulti = mainStatMulti(stats.levelStats, stats.jobStats, stats[stats.jobStats.autoAttackStat]);
```
xiashtra, line 36: Won't this cause casters and healers to have their auto-attacks buffed by intelligence/mind potions?  Since their auto-attacks use strength for damage calculations, they should not get buffed by their mainStat potions.

xpdota: It shouldn't, because it still looks up `stats[stats.jobStats.autoAttackStat]`, which wouldn't have been modified. This is just telling it to recalculate the multipliers in the same way that they would be originally calculated.

I plan to eventually replace this logic by making a full class for ComputedSetStats which can handle all of this logic internally.

packages/core/src/sims/common/potion.ts
```diff
@@ -0,0 +1,47 @@
...
+        name: `Grade ${grade} Tincture of ${statName}`,
+        // Does it make the most sense to have this be the ability ID + item ID?
+        // From a sim standpoint, there's no reason to care about the potion's action ID, only
+        // the item ID and buff ID.
+        // For now, going with the same thing you'd see in an ACT log
+        id: id,
+        itemId: itemId,
+        type: 'ogcd',
+        attackType: 'Item',
+        animationLock: 1.5,
```
xiashtra, line 23: Do you have a source for 1.5 lock on potions?  My understanding is potions are 1.1 (and food may be slightly longer).

xpdota: I'll open another PR to change it to 1.1

PR #137, opened 2024-06-16
Shared cooldown support

xpdota:
Adds `sharesCooldownWith` field to `Cooldown`. If you want A, B, C, and D to have shared cooldowns, then B/C/D should have `sharesCooldownWith: A`.

The source and target cooldowns can have different CDs. If one has a 60s CD, and one has a 120s CD, then the cooldown time will be derived from whichever you actually use. That is, using the first skill would block both skills for 60s, and using the second would block both skills for 120s.

Behavior is undefined for abilities with more than one charge.

### DIFF
--- a/packages/core/src/sims/common/cooldown_manager.ts
+++ b/packages/core/src/sims/common/cooldown_manager.ts
@@ -1,4 +1,4 @@
-import {Ability} from "../sim_types";
+import {Ability, CdAbility} from "../sim_types";
 
 export type CooldownMode = 'none' | 'warn' | 'delay' | 'reject';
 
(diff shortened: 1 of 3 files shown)


xiashtra:
> Adds `sharesCooldownWith` field to `Cooldown`. If you want A, B, C, and D to have shared cooldowns, then B/C/D should have `sharesCooldownWith: A`.

Just to clarify, the intention is that if A and B share a cooldown, B should reference A via `sharesCooldownWith: A`, but A *should not* reference B via `sharesCooldownWith: B`, because it would cause a circular-reference?  This seems like an easy trap to fall into for sim implementers.  Perhaps there is a better way to implement this that would prevent creating a footgun.

PR #144, opened 2024-06-20
Implement caster variant of damage formula

xpdota:
Also updates tests, and implements the 'flp' function that floors to a certain precision. 

e.g. `flp(3, value)` will floor to 3 places, i.e. the equivalent of `fl(value * 1000) / 1000`.


xiashtra:
Two notes:
- caster/healer autos use the same auto-attack formula as physical jobs (essentially the physical damage formula)
- BLU has two special abilities that use the physical formula despite being magical actions (these also happen to be the channeled actions that use a different tick rate as well)
I would not be concerned if the BLU actions can't be properly supported for now.

xpdota:
Ok, I fixed the first one. I'll say file a bug for the second and I'll see what I can do.

PR #149, opened 2024-06-22
Ninja Sim

Makar8000:
~~depends on #143~~

TODO:
- [x] **Fix burst alignment.**
- [x] **Optimize gauge usage.**
- [x] Clean up filler oGCD usage (currently is based on a counter rather than actually checking if an ability will clip or not).
- [x] Update IDs/Potencies once Dawntrail is available.
- [x] Clean up `getRotationsToSimulate` to use provided cycle options instead of wrapping it in a while loop
- [x] Add a function to display gauge data in the rotation output


packages/frontend/src/scripts/sims/melee/nin/nin_lv100_sim.ts
```diff
@@ -0,0 +1,415 @@
...
+
+        this.useNinjutsu(cp, Actions.Raiton);
+        this.useFillerOgcd(cp, 1, 50);
+
+        this.useCombo(cp, true);
+    }
+
+    getRotationsToSimulate(): Rotation[] {
+        const sim = this;
+        this.rotationState = new RotationState();
```
xpdota, line 373: You can also subclass CycleProcessor and add custom behavior there. That avoids needing to reset state and having to pass the cycleProcessor around to all the extra methods.

packages/core/src/sims/buffs.ts
```diff
@@ -1,8 +1,8 @@
 import {Ability, PartyBuff} from "./sim_types";
 
 
-export const Mug = {
-    name: "Mug",
+export const Dokumori = {
```
xpdota, line 4: The name is used as the key for enabling/disabling the buff, so I think I might need to add logic here to not break existing sim exports.

Makar8000: Where is it used? I just did a find and replace for it

packages/frontend/src/scripts/sims/melee/nin/nin_lv100_sim.ts
Makar8000: I'll update this tonight

Makar8000:
added a todo list to OP to be completed before I remove this from draft state

packages/core/src/sims/buffs.ts
xpdota: It's used by BuffSettingsManager to import/export party buff settings for sims. So it could cause unexpected results for existing sheets. I'd say the best solution would be to add a field to override the save key.

Makar8000: resolved in #151

packages/core/src/sims/cycle_sim.ts
```diff
@@ -475,15 +475,33 @@ export class CycleProcessor {
...
         });
     }
 
+    /**
+     * Modifies the stack value for a given buff. The stack value provided should be the modified amount and not the final amount
+     * 
+     * @param buff The Buff
+     * @param stacks The stack modification to add
+     */
+    modifyBuffStacks(buff: Buff, stacks: number) {
```
xpdota, line 491: Having it treat 'stacks' as a delta rather than the final amount could be confusing. I'd suggest changing the parameter name to something like 'stackDelta'. 

Better yet, maybe have modifyBuffStacks just take the final amount, and then separate `addBuffStacks()` and `subtractBuffStacks()`.

xpdota: Or, have it take a number => number function.

Makar8000: The main reason why I did this is because the current implementations of beforeSnapshot, beforeAbility, etc, do not have access to the current buff itself. So there is no way to actually know what the new total is in it's current implementation. I can rename the function, or just remove it entirely and use my custom cycle processor instead. 

Would you rather I:
- Just rename the param
- Rename the param, make it private, and add a `addBuffStacks(stacks: number)` and `subtractBuffStacks(stacks: number)` instead.
- Remove it all completely and just update the stacks in my own Cycle Processor
?

xpdota: I think #2 but it's fine to keep it public in case there's a use for it.

PR #151, opened 2024-06-23
Add saveKey for buffs

Makar8000:


packages/core/src/sims/buffs.ts
```diff
@@ -249,4 +265,4 @@ export const ALL_BUFFS = [
     AstCard, OffGuardBuff
 ] as const;
 
-export type BuffName = typeof ALL_BUFFS[number]['name'];
+export type BuffName = typeof ALL_BUFFS[number]['name'] & typeof ALL_BUFFS[number]['saveKey'];
```
xpdota, line 268: What is the intent of this? This would cause it to exclude anything that has a different name vs save key. 

I think it might be better to just have a separate type for BuffName vs BuffSaveKey?

Makar8000: Oops, I meant to use `|` instead of `&`. The point is so that this type represents whatever the save key is, which defaults to name if there is no save key.

Makar8000: fixed and renamed to a more appropriate name

xpdota:
Thank you!

PR #182, opened 2024-07-03
Paladin Skill Speed Simulator

Chromatophore:
Hello, I have made a paladin simulator that aims to demonstrate the limited but attainable benefits of skill speed via effectively executing a variety of strategies.

I've marked it as a non default sim for PLD as this a) isn't an infinite dummy model, and b) PLD is going to be a 2.50 GCD job as soon as it can be - as such, Nikroulah's PLD cycle sim is a much better tool to select a BIS gearset. However, it's still quite easy to add this one, if desired, and ensures that people see its copious description. It is intended more of a playground to test the impact of non-optimal behaviors, as well as demonstrate possible approaches to situations outside of the norm - showing the impact of mistakes as much as it aims to show any benefits.

It is not abnormal to find the Skill Speed stat on ones gear, especially early in tiers or during prog. Without a simulator like this, it can be very difficult to gauge between 2 close in ilvl items, where the higher one has sks - it may be desirable to a Paladin for it's other stats, or, it may be more trouble that it is worth to them. Being able to get a feel for it's benefit, while also taking into consideration the methods they might use to engage with the substat (including to pretend it does not exist) seems quite pragmatic to me.

### DIFF
--- a/packages/frontend/src/scripts/sims/tank/pld/pldsks_sim.ts
+++ b/packages/frontend/src/scripts/sims/tank/pld/pldsks_sim.ts
@@ -661,7 +661,8 @@ export class PldSKSSheetSim extends BaseMultiCycleSim<PldSKSSheetSimResult, PldS
 						{
 							safety++;
 							// Note the current status:
-							cp.addSpecialRow(`FOF: ${fof_remaining}, melee+1? ${enough_time_if_melee}`); 
+							if (!strategy_250)
+								cp.addSpecialRow(`FOF: ${fof_remaining}, melee+1? ${enough_time_if_melee}`); 
 							// Use a burst GCD
 							cp.useBurstFiller(prioritise_melee);
 							// Use any remaining oGCDs:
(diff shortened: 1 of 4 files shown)


Chromatophore:
I have, perhaps, appeased the linter.

packages/frontend/src/scripts/sims/tank/pld/pldsks_sim.ts
```diff
@@ -0,0 +1,863 @@
...
+	debugState() {
+		console.log( [ this.combo_state , this.sword_oath , this.divine_might ].toString() );
+	}
+}
+
+export interface PldSKSSheetSimResult extends CycleSimResult {
+}
+
+export interface PldSKSSheetSettings extends SimSettings {
+	acknowledgeSKS: boolean,
```
xpdota, line 97: I think there's some mismatched indentation

Chromatophore: Should, hopefully, be resolved. Half the file was on tabs, half was on spaces, should all be spaces now. Linter appeased locally so should be happy here, also.

PR #187, opened 2024-07-06
Rpr sim update for dawntrail

Junie-774:
Two major changes:
- updated RPR to be cleaner and also updated the rotation for dawntrail
- added support for buffs which have stacking duration (applicable to RPR, VPR, DNC, etc.)

Needs RPR mentor review before getting pushed to live


packages/frontend/src/scripts/sims/melee/rpr/rpr_sheet_sim.ts
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    getRotationsToSimulate(): Rotation<RprCycleProcessor>[] {
+        this.rotationState = new RotationState();
+        return [{
+            cycleTime: 120,
+
+            /* I just used the cycle processor instead of doing any cycles, as doing so
+             * would require me to duplicate all of the 'useX(cp)' functions for CycleContext
+             * Is there a better way to do this?
```
xpdota, line 377: You don't have to duplicate any methods. You can still access the CycleProcessor within a cycle block.

Junie-774: oh, teehee, that was an old comment from the original implementation

xpdota:
Thank you!

PR #193, opened 2024-07-11
Fix DNC auto usages

nikroulah:
auto potency based on auto delay is actually baked into the auto dmg calculation, so the number of usages should be based on the actual number of autos rather than the normalized p/3s "usages"

### DIFF
--- a/packages/frontend/src/scripts/sims/ranged/dnc_sim.ts
+++ b/packages/frontend/src/scripts/sims/ranged/dnc_sim.ts
@@ -453,7 +453,7 @@ export class DncDtSim extends BaseUsageCountSim<DncDtSimResults, DncDtSimSetting
             return []
         }
         result.push(...common);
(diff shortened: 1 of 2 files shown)


packages/frontend/src/scripts/sims/ranged/dnc_sim.ts
```diff
@@ -453,7 +453,7 @@ export class DncDtSim extends BaseUsageCountSim<DncDtSimResults, DncDtSimSetting
             return []
         }
         result.push(...common);
-        result.push([auto, (buffDuration ?? FullCycleTime) / 3]);
+        result.push([auto, (buffDuration ?? FullCycleTime) / 3.12]);
```
xpdota, line 456:
```suggestion
        result.push([auto, (buffDuration ?? FullCycleTime) / set.computedStats.aaDelay]);
```

PR #222, opened 2024-07-24
Reworked AC & pot logic

Junie-774:
- Reworked AC logic, was previously bad
- Fixed sim crashes (hopefully)
- Reworked pot logic to be better


xpdota:
Thanks!

PR #244, opened 2024-07-29
Adjust potion exports

Makar8000:
This PR Adjusts the potion exports so that there is one generic "max level" potion. This way, sims that only care to use the max grade potion available do not need to be updated every time the game adds a new potion.

I have also exported a few extra variables, so that sims could add settings in the future that would allow users to configure which grade potion they would like to use.


xpdota:
Thanks!

PR #252, opened 2024-07-31
Paladin Skill Speed Sim Update 7.05

Chromatophore:
I've updated my pld sim quite a lot, for both 7.05, as well as adding a variety of additional options which are becoming quite pertinent. I've also marked myself as the maintainer, and shortened the description, as it no longer fit in the information box when adding the sim.

I also renamed my actions/buff files to have sks in the name to help differentiate them from Nik's.


packages/frontend/src/scripts/sims/tank/pld/pldsks_sim.ts
```diff
@@ -351,11 +442,57 @@ export class PldSKSSheetSim extends BaseMultiCycleSim<PldSKSSheetSimResult, PldS
                 const physGCD = cp.stats.gcdPhys(2.5);
                 const magicGCD = cp.stats.gcdMag(2.5);
 
+                const GCDActionSet = new Map<string, GcdAbility>();
+                const oGCDActionSet = new Map<string, OgcdAbility>();
+
+                GCDActionSet.set("Fast",Actions.FastBlade);
```
xpdota, line 448: I'm slightly worried about this design, since the keys here don't receive any type checking when you use them as it's just a `string` type.

xpdota:
Thanks!

PR #257, opened 2024-08-03
Fix dragging hand getting stuck on

jjspace:
The cursor style applied while dragging was not getting removed causing the grabbing hand to always be applied.

### Test

1. Create or use a sheet with more than 2 gearsets
2. Drag one of the sets
3. Ensure the grabbing hand does not remain stuck on after dropping the gearset


xpdota:
Thanks!

PR #280, opened 2024-08-15
Add button to unequip item

### LINKED ISSUE #253: Unable to de-select Items
There is currently no ability to select "none" for a slot or food.
Would be nice to either have a "none" dummy option

xpdota:
image

Closes #253


Makar8000:
Not a fan of the design/placement if I'm being honest, but at least it's something for now

xpdota:
> Not a fan of the design/placement if I'm being honest, but at least it's something for now

Do you have a suggestion? I was thinking of putting it at the top of the box, but I went with this design since I think there's some UX value in having it on the specific item you're unequipping.

Makar8000:
I don't have any specific suggestion, but as-is it's odd for the delete icon to be in the middle of the row. I agree at the top would be worse. What about making it so that when you click on an already selected item, it just unselects it (really need that undo or materia save feature though). Another option is at the start or end of a row (not start/end of item name)

PR #295, opened 2024-08-26
Fix for Pld SKS Sim near Infinite Loop

Chromatophore:
If the phase was set to end at a particular point, the sim would nearly infinite loop as it found it could fit an 100k+ GCDs into FOF, which would be far more SKS than anyone could reasonably meld...

Since the GCDs took place after time had ended, the sim still calculated DPS correctly. However, opening the sim's table of events would cause the browser to go OOM as it tried to also render 100k comments.


Chromatophore:
NB This will lock up your browser tab for a minute, and probably crash the tab in the unpatched version:

Here's an example exported sheet which causes the bug to exhibit on the 1st and 3rd SKS sim, but not the 2nd, due to the specifics of their settings (click the DPS number to try and view the table, and subsequently blow up your tab)

https://xivgear.app/?page=sl%7C133ca913-79b9-4dee-9863-bc5596396660

I've also tested sim durations from 300-400 to ensure that no other times have a similar problem with this fix in place.

Issue #307, opened 2024-09-04
Materia levels are not respected correctly


xpdota:
Currently, it is possible to equip higher-level materia onto lower-level items.

The materia filtering is currently working off the level of the *sheet*, not the level of the *item*. The latter is the correct behavior.

PR #320, opened 2024-09-12
Make materia selection respect ilvl

xpdota:
Currently, maximum materia level is only determined by sheet level. Instead, it should be determined by the ilvl of the item it is being equipped into. This affects both manual picking and auto-fill.


xpdota:
Closes #307

PR #327, opened 2024-09-18
Split rotations and damage new

Junie-774:
Splits up the rotation-generating step and the calculating-damage step of `BaseMultiCycleSim` and `CycleProcessor`.

`BaseMultiCycleSim` now caches the most recent rotation and only re-generates it when speed stat changes, instead of generating the rotation with every call to `simulate()


xpdota:
Thanks!

PR #335, opened 2024-09-26
Update dnc sim

Junie-774:
Updates the DNC sim for DT release
- updates potencies
- updated some usage logic

This is intended to reflect the calculations in the DNC gear planner


xpdota:
Thanks!

PR #336, opened 2024-09-27
Split sim logic from UI

Junie-774:
Splits off Simulation from its UI elements, and moves simulation logic into `core`.

Every sim has been investigated and seems to work fine


packages/frontend/src/scripts/sims/blu/blue_common_ui.ts
```diff
@@ -0,0 +1,22 @@
+import { FieldBoundCheckBox, labeledCheckbox } from "@xivgear/common-ui/components/util";
```
xiashtra, line 1: this file should be `blu_common_ui.ts` to keep the naming convention consistent.

packages/frontend/src/scripts/test/sims/cycle_processor_tests.ts
```diff
@@ -24,7 +24,7 @@ import {
     SimSpec
 } from "@xivgear/core/sims/sim_types";
 import {CycleProcessor, CycleSimResult, ExternalCycleSettings, Rotation} from "@xivgear/core/sims/cycle_sim";
-import {BaseMultiCycleSim} from "../../sims/sim_processors";
+import { BaseMultiCycleSim } from '@xivgear/core/sims/processors/sim_processors';
```
xpdota, line 27: If these tests don't involve the GUI at all, they can be moved to the core module as well.

packages/frontend/src/scripts/sims/simulation_gui.ts
```diff
@@ -0,0 +1,76 @@
+import { CharacterGearSet } from "@xivgear/core/gear";
+import { SimResult, SimSettings, SimSpec, Simulation } from "@xivgear/core/sims/sim_types";
+
+export type ResultTypeOfSim<Sim> = Sim extends Simulation<infer R, unknown, unknown> ? R : never;
+export type SettingsTypeOfSim<Sim> = Sim extends Simulation<SimResult, infer S, unknown> ? S : never;
+export type ExportSettingsTypeOfSim<Sim> = Sim extends Simulation<SimResult, unknown, infer SE> ? SE : never;
+
+export abstract class SimulationGui<ResultType extends SimResult, SettingsType extends SimSettings, SettingsExport>
+    implements Simulation<ResultType, SettingsType, SettingsExport> {
```
xpdota, line 9: Does this need to implement Simulation? I was imagining that the `SimulationGui` object would only have the GUI-specific parts, and everything else deals with the raw `Simulation`.

xpdota: What I was thinking is that the generic type of `SimulationGui` would be the sim type itself.

Junie-774: yea u rite, removing

packages/frontend/src/scripts/test/sims/cycle_processor_tests.ts
Junie-774: ooh yeah!

packages/frontend/src/scripts/sims/simulation_gui.ts
xpdota: (This also avoids the need for the `declare` workaround)

Junie-774: > What I was thinking is that the generic type of `SimulationGui` would be the sim type itself.

I tried to do this; but stuff got screwy when i tried to implement BaseMultiCycleSimGui

Junie-774: > Does this need to implement Simulation? I was imagining that the `SimulationGui` object would only have the GUI-specific parts, and everything else deals with the raw `Simulation`.

oh ignore the commit for this, more stuff breaks when you actually remove the functions rather than just "implements"

xpdota:
Thanks!

PR #341, opened 2024-10-07
Add DRK Sim support

VioletHynes:
Added a DRK sim, and much of the scaffolding for future work.

Currently, it only supports level 100, and only fully supports 2.50 and 2.46 (the recommended speeds for level 100 DRK). Future work will include making a usage based option that can support generic GCDs.

MP isn't perfect, but it's mostly there (it's more pessimistic than reality), and I've not yet added the code for pre-pull TBN, but blood and the sim in general works fully:
image
image

Future work for MP may include making it core for every job, or something similar, but this is more than fine for now. 

The only other intricacy is that Living Shadow is programmed simply as one big 2450 potency chunk, which is not correct. However, doing that seemed good enough for now as a first draft.

It does work as expected, and has a default sim time of 8m 30s:
image

As far as I can tell, all of the numbers look realistic.

There's more to do, but I didn't want perfect to be the enemy of good, and getting it merged/approved opens the way for other contributors. I've been running this by the DRK mentor team at The Balance and we agree it's in a good enough spot to merge.

Happy to address feedback/make changes as necessary.


packages/core/src/sims/tank/drk/drk_types.ts
```diff
@@ -0,0 +1,129 @@
...
+    updateBloodGauge?(gauge: DrkGauge): void;
+    
+    /** The Blood cost of the ability */
+    bloodCost?: number;
+
+    /** Run if an ability needs to update MP */
+    updateMP?(gauge: DrkGauge): void;
+    
+    /** The MP cost of the ability */
+    mp?: number;
```
amakiirs, line 17: Arguably, this could be combined into the `updateMP` function, and similarly for blood. I think that's broadly a good idea, but not super important for now, and also may become obsolete if gauge becomes generic as has been discussed in the past.

packages/core/src/sims/tank/drk/drk_lv100_sim.ts
```diff
@@ -0,0 +1,258 @@
...
+
+export interface DrkSimResult extends CycleSimResult {
+
+}
+
+export interface DrkSettings extends SimSettings {
+    usePotion: boolean;
+    prepullUnmend: number; //the number given is how many seconds prepull we use Unmend.
+    fightTime: number; // the length of the fight in seconds
+    // TODO: should we add a prepull Shadowstride option? Would be nice to compare on differing killtimes
```
amakiirs, line 22: I think that's probably largely irrelevant for the core gear comparison purpose, but useful when rotation generation is added. For this sim, it seems unimportant.

amakiirs (approved):
Finally double checked everything, and it all still looks reasonable. MP might need different handling in the future (e.g., for rotation generation), but what there currently is seems perfectly serviceable even for the starting point of that (though it's missing prepull MP, as far as I can tell). Nothing else really stood out that wasn't already noted as a limitation of this sim, and I think it makes clear to the user those limitations, as far as it's relevant, anyway.

packages/core/src/sims/tank/drk/drk_lv100_sim.ts
VioletHynes: Yeah -- the main advantage for something like this in my view is when using the sim for comparing numbers at differing kill times. It certainly isn't the most important thing to address (I think my next priority is Living Shadow -- though I've been trying not to get too ahead of myself before this is approved and merged) but it'd be a nice thing to do at some point.

PR #343, opened 2024-10-09
Fix the mouseover info in the stat tier display

KekMaster:
An accidental "this" appears in stat_tier_display.ts, which overrides the main stat tier div instead of the expansion DIVs
This is inside the loop that generates the different remove/add materia variations for the stat tiers

It causes 2 issues:
1. The popup info for the lowerLeftDiv is wrong
2. There's no popup info for the any of the expansion lowerLeftDivs

### This is the "wrong info" issue:
image

### For reference, this is from the /math page
image

### This is how it should look like:
image

Removing the "this" fixes both of those issues


xpdota:
Thanks!

PR #345, opened 2024-10-11
Add some extra sims

wolfcomp:
## Sims being added:
- DRG TOP

## Sims to be worked on:
- MNK TOP
- MNK 100 DL 5s
- MNK 100 DL 7s
- MNK 100 SL 5s
- MNK 100 SL 7s

If preferable, I can split these into separate PRs

### DIFF
--- a/packages/core/src/sims/melee/drg/drg_actions.ts
+++ b/packages/core/src/sims/melee/drg/drg_actions.ts
@@ -1,0 +1,217 @@
+import { DiveReady, LanceCharge, LifeOfTheDragon, LifeSurge, NastrondReady, PowerSurge } from './drg_buffs';
+import { Litany } from '@xivgear/core/sims/buffs'
+import DRGGauge from './drg_gauge';
+import { DrgGcdAbility, DrgOgcdAbility } from './drg_types';
+
+export const TrueThrust: DrgGcdAbility = {
+    type: 'gcd',
+    name: "True Thrust",
+    id: 75,
+    attackType: "Weaponskill",
+    potency: 230,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const VorpalThrust: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Vorpal Thrust",
+    id: 78,
+    attackType: "Weaponskill",
+    potency: 280,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const HeavensThrust: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Heavens' Thrust",
+    id: 25771,
+    attackType: "Weaponskill",
+    potency: 400,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const FangAndClaw: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Fang and Claw",
+    id: 3554,
+    attackType: "Weaponskill",
+    potency: 300,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const Drakesbane: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Drakesbane",
+    id: 36952,
+    attackType: "Weaponskill",
+    potency: 400,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const RaidenThrust: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Raiden Thrust",
+    id: 16479,
+    attackType: "Weaponskill",
+    potency: 280,
+    gcd: 2.5,
+    cast: 0,
+    updateGauge: (gauge: DRGGauge) => {
+        gauge.addFirstmindsFocus();
+    }
+};
+
+export const Disembowel: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Disembowel",
+    id: 87,
+    attackType: "Weaponskill",
+    potency: 250,
+    gcd: 2.5,
+    cast: 0,
+    activatesBuffs: [PowerSurge]
+};
+
+export const ChaoticSpring: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Chaotic Spring",
+    id: 25772,
+    attackType: "Weaponskill",
+    potency: 300,
+    gcd: 2.5,
+    cast: 0,
+    dot: {
+        tickPotency: 45,
+        duration: 24,
+        id: 0
+    }
+};
+
+export const WheelingThrust: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Wheeling Thrust",
+    id: 3556,
+    attackType: "Weaponskill",
+    potency: 300,
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const LifeSurgeAction: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Life Surge",
+    id: 83,
+    potency: 0,
+    cooldown: {
+        time: 40,
+        charges: 2
+    },
+    attackType: 'Ability',
+    activatesBuffs: [LifeSurge]
+};
+
+export const LanceChargeAction: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Lance Charge",
+    id: 85,
+    potency: 0,
+    cooldown: {
+        time: 60
+    },
+    attackType: 'Ability',
+    activatesBuffs: [LanceCharge]
+};
+
+export const DragonfireDive: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Dragonfire Dive",
+    id: 96,
+    potency: 500,
+    cooldown: {
+        time: 120
+    },
+    attackType: 'Ability',
+};
+
+export const BattleLitany: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Battle Litany",
+    id: 3557,
+    potency: 0,
+    cooldown: {
+        time: 120
+    },
+    attackType: 'Ability',
+    activatesBuffs: [Litany]
+};
+
+export const Geirskogul: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Geirskogul",
+    id: 3555,
+    potency: 280,
+    cooldown: {
+        time: 60
+    },
+    attackType: 'Ability',
+    activatesBuffs: [NastrondReady]
+};
+
+export const Nastrond: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Nastrond",
+    id: 7400,
+    potency: 360,
+    attackType: 'Ability',
+    cooldown: {
+        time: 2
+    }
+};
+
+export const MirageDive: DrgOgcdAbility = {
+    type: 'ogcd',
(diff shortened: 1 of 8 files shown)


packages/core/src/sims/melee/drg/drg_top_sim.ts
```diff
@@ -0,0 +1,299 @@
...
+    stub: "drg-top-sim",
+    displayName: "DRG Ultimate LvL 90 Sim",
+    makeNewSimInstance: function (): DrgTopSim {
+        return new DrgTopSim();
+    },
+    loadSavedSimInstance: function (exported: DRGTopSettingsExternal): DrgTopSim {
+        return new DrgTopSim(exported);
+    },
+    supportedJobs: ["DRG"],
+    supportedLevels: [100],
```
xpdota, line 33: If this is a level 90 sim, then the supported level should be 90.

packages/core/src/sims/melee/drg/drg_top_sim.ts
```diff
@@ -0,0 +1,299 @@
...
+    isDefaultSim: false,
+    maintainers: [{
+        name: "WildWolf",
+        contact: [{
+            type: "discord",
+            discordTag: "wildwolf",
+            discordUid: "108974906472955904"
```
xpdota, line 40: This should be the user-provided tag, rather than the uid number.

wolfcomp: and for the `discordUid` some of the others I saw had it as the `User ID` so not sure what you mean is wrong here

packages/core/src/sims/melee/drg/drg_top_sim.ts
wolfcomp:
```suggestion
    supportedLevels: [90],
```

^ should use that feature

packages/core/src/sims/melee/drg/drg_top_sim.ts
xpdota: Oh, nevermind, I had it confused.

wolfcomp:
If it complains about linting, I have no idea how to get the same linter to run on my machine to actually fix it as this is the result i get

wolfcomp:
finally got the linter to show the same as the workflow so now this should be happy

packages/core/src/sims/melee/drg/drg_top_sim.ts
```diff
@@ -0,0 +1,300 @@
...
+    stub: "drg-top-sim",
+    displayName: "DRG Ultimate LvL 90 Sim",
+    makeNewSimInstance: function (): DrgTopSim {
+        return new DrgTopSim();
+    },
+    loadSavedSimInstance: function (exported: DRGTopSettingsExternal): DrgTopSim {
+        return new DrgTopSim(exported);
+    },
+    supportedJobs: ["DRG"],
+    supportedLevels: [90],
```
VioletHynes, line 33: There's probably not a lot of extra work to get this supported for level 100, too, though I understand that is extra work.

VioletHynes: If we did this, we wouldn't need to call it `drg_top_sim` either, it'd just be `drg_sheet_sim`. If you're interested in an example, DRK supports 70-100.

packages/core/src/sims/melee/drg/drg_actions.ts
```diff
@@ -0,0 +1,217 @@
...
+    gcd: 2.5,
+    cast: 0,
+};
+
+export const Drakesbane: DrgGcdAbility = {
+    type: 'gcd',
+    name: "Drakesbane",
+    id: 36952,
+    attackType: "Weaponskill",
+    potency: 400,
```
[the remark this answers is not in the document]
wolfcomp: I would say that this is something that should be done if the other sims gets implemented and not tossed to this portion.

packages/core/src/sims/melee/drg/drg_top_sim.ts
```diff
@@ -0,0 +1,300 @@
...
+    isDefaultSim: false,
```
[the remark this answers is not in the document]
wolfcomp: I would say only the end level sim should be the default which would be in this case 100

packages/core/src/sims/melee/drg/drg_top_sim.ts
```diff
@@ -0,0 +1,300 @@
...
+
+export interface DRGTopSimResult extends CycleSimResult {
+
+}
+
+export interface DRGTopSettingsExternal extends ExternalCycleSettings<DRGTopSimSettings> {
+
+}
+
+export const drgTopSimSpec: SimSpec<DrgTopSim, DRGTopSettingsExternal> = {
```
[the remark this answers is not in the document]
wolfcomp: This is something i can change when fixing the merge conflicts

packages/core/src/sims/melee/drg/drg_top_sim.ts
wolfcomp: I would rather not have to maintain an Excel sheet thing in order for this to work. I know that it is something that could be done, and more me just not wanting to rely on other services working for a single sim to work.

packages/core/src/sims/melee/drg/drg_top_sim.ts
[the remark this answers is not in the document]
VioletHynes: A default sim will show up only for its supported levels, so you should be safe here. If setting this to true makes the sim show up for level 100 gearsets, then there'd be a bug

packages/core/src/sims/melee/drg/drg_top_sim.ts
VioletHynes: Perhaps the nomenclature is confusing here, but "sheet" in this comment doesn't mean an Excel sheet, it means gearsheet. This wouldn't mean needing to support any other service, it's just a naming thing. Most of our sims are called job_sheet_sim etc.

packages/core/src/sims/melee/drg/drg_actions.ts
```diff
@@ -0,0 +1,217 @@
...
+    },
+    attackType: 'Ability',
+    activatesBuffs: [NastrondReady, LifeOfTheDragon],
+};
+
+export const Nastrond: DrgOgcdAbility = {
+    type: 'ogcd',
+    name: "Nastrond",
+    id: 7400,
+    potency: 360,
```
VioletHynes, line 169: This (and Nastrond Ready) will need to be updated following the balance patch.

packages/frontend/src/scripts/sims/melee/drg/drg_90_sim_ui.ts
```diff
@@ -0,0 +1,67 @@
...
+                    div.style.display = 'flex';
+                    div.style.alignItems = 'center';
+                    div.style.gap = '6px';
+                    div.style.padding = '2px 0 2px 0';
+                    div.style.boxSizing = 'border-box';
+
+                    const span = document.createElement('span');
+                    span.textContent = `${firstmindsFocus}`;
+
+                    const barOuter = document.createElement('div');
```
VioletHynes, line 31: nit: instead of a bar, we could consider using circles to note stacks, like the Monk Sim for Chakras or the GNB sim for carts

packages/core/src/sims/melee/drg/drg_buffs.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export const LifeOfTheDragon: Buff = {
+    name: "Life of the Dragon",
+    selfOnly: true,
+    effects: {
+        dmgIncrease: 0.15,
+    },
+    duration: 20,
+    // No status ID due to being a hidden buff
```
VioletHynes, line 31: Since it's hidden, it would be great to show in the UI

packages/core/src/sims/melee/drg/drg_90_sim.ts
```diff
@@ -0,0 +1,300 @@
...
+            hideCycleDividers: true,
+        });
+    }
+    getRotationsToSimulate(set: CharacterGearSet): Rotation<DRG90CycleProcessor>[] {
+        return [{
+            cycleTime: 120,
+            apply(cp: DRG90CycleProcessor) {
+                const numPots = Math.ceil(cp.remainingTime / 360); // Force pots during 2mins
+                const maxAllowableDelay = cp.remainingTime - ((numPots - 1) * 360);
+                const potOpener = maxAllowableDelay <= 120;
```
VioletHynes, line 287: The pot doesn't seem to be applying in the opener from the image, and the above logic seems like it would need to be in the remainingCycles function somewhere (e.g. `useOgcdCooldown`) to actually use pots past the opener

RhyTamok:
This rotation is not ready to ship. Judging by the screenshot there are multiple major (and minor) issues that need to be addressed first.

- Rotation is started with the wrong combo. Always needs to be Chaotic Spring combo first to apply Power Surge.
- Every combo is broken. The respective second combo actions (Disembowel and Vorpal Thrust) are always replaced by a True Thrust. This also means that Power Surge is never accounted for.
- Logic around double weaves needs to be adjusted. There should not be any double weaves with Stardiver
- Logic for Life Surge has to be adjusted. Unless we're talking about the last 0.1% of optimization it should only be used on Heavens' Thrust and Drakesbane. It is used on Chaotic Spring in the opener here. It should also be held for burst windows and not necessarily used on cooldown
- Geirskoguls 15% damage buff is not accounted for
- Stardiver is used outside of Life at 42s
- This isn't clear from the screenshot but Wyrmwind Thrust should also be held for buffs if possible (Always use before next Raiden, hold for any buffs applied before that)

The rotations for level 90 and 100 are virtually identical. You only gain 2 more follow-up skills and a replacement for the second combo skills, so as long as the double weave logic is sounds, it should be viable for both.
I am not sure how skillspeed is handled here but for speeds faster than 2.50 it becomes considerably more complex as you have to account for 9 GCD buff windows and delaying oGCD to not clip while still keeping optimal buff windows.

packages/core/src/sims/melee/drg/drg_buffs.ts
wolfcomp: It both is hidden and not hidden. The reason being that the status itself isn't a displayed thing, but it does show it through the Job Gauge with a countdown timer

packages/core/src/sims/melee/drg/drg_90_sim.ts
wolfcomp: Then most likely another sim is also broken since I took most of the pot logic from there

packages/core/src/sims/melee/drg/drg_buffs.ts
RhyTamok: Life of the Dragons 15% damage up still has to be shown in the total buffs column and it should be listed with an icon to make it more readable. While there is no direct buff icon for it in game, the status icon for LotD still exists https://arcanedisgea.github.io/xiv-id/?index=Status&query=life+of+the+dragon
We always display it as an active buff in our rotation builders so people are used to seeing it.

PR #358, opened 2024-10-16
[DRK Sim] Computationally generate rotation for DRK sim

VioletHynes:
Update the DRK sim to have a computationally generated rotation. It seems to fully work, with no mana overcap.

There are some triple weaves, but realistically this would only impact a very long fight and on certain speeds. Personally, I think it's fine for now.

Works for all GCDs.

image

(Numbers seem higher than old sim because pots work now and I believe the GCD pro-rating also works better with this approach)


xpdota:
Thanks!

PR #364, opened 2024-10-18
[DRK Sim] Living Shadow/weird killtime bug fix

VioletHynes:
If running multiple sims or changing between two killtimes, and one of those killtimes is before Living Shadow's full abilities went off, it might apply those abilities to the next simulation, which is obviously wrong. This fixes that issue.

Sorry for missing this -- Living Shadow is inherently weird.


xpdota:
The ideal way to fix this would be to move any mutable state to the custom CycleProcessor, since that gets one instance per sim invocation and thus can't leak data from one run to another.

VioletHynes:
@xpdota thanks -- fixed

PR #366, opened 2024-10-18
[WAR Sim] Warrior Sim Initial Implementation

VioletHynes:
An initial implementation of a Warrior Sim. Supports level 100.

Forced Crit + DHits needed to be fixed. They were making crits applying twice. I removed the application from the part where crits are not normally applied.

I needed to add some additional support for the cooldown reduction of Infuriate and buff extension for Inner Release.

Numbers seem accurate:
image

Everything seems to work as expected:
image


packages/core/src/sims/sim_utils.ts
```diff
@@ -32,6 +36,10 @@ function potencyToDamage(stats: ComputedSetStats, potency: number, dmgAbility: D
         }
     });
     const nonCritDmg = baseDamageFull(modifiedStats, potency, dmgAbility.attackType, forceDhit, forceCrit);
+    if (forceDhit && forceCrit) {
```
xiashtra, line 39: I think this might cause a problem with actions that are either an autocrit or autodhit, but not both (SAM comes to mind).

VioletHynes: Yeah, that's a good point -- I was mostly just doing this to test if this fixed it, but it's looking like I'll need a deeper look regardless. Thanks for pointing this out though, it's a case I'll need to look out for.

packages/xivmath/src/xivmath.ts
```diff
@@ -272,7 +274,7 @@ function usesCasterDamageFormula(stats: ComputedSetStats, attackType: AttackType
 /**
  * Computes base damage. Does not factor in crit/dh RNG nor damage variance.
  */
-export function baseDamageFull(stats: ComputedSetStats, potency: number, attackType: AttackType = 'Unknown', autoDH: boolean = false, autoCrit: boolean = false, isDot: boolean = false): ValueWithDev {
+export function baseDamageFull(stats: ComputedSetStats, potency: number, attackType: AttackType = 'Unknown', autoDH: boolean = false, isDot: boolean = false): ValueWithDev {
```
VioletHynes, line 274: This function was extraneously applying crit/dh for autocrits/autodhs, so it was removed from here. The multipliers are applied where they are for normal crit/dhs (`applyDhCritFull`).

packages/xivmath/src/xivmath.ts
```diff
@@ -423,15 +407,15 @@ export function applyDhCrit(baseDamage: number, stats: ComputedSetStats) {
     return baseDamage * (1 + stats.dhitChance * (stats.dhitMulti - 1)) * (1 + stats.critChance * (stats.critMulti - 1));
 }
 
-export function dhCritPercentStdDev(stats: ComputedSetStats) {
+export function dhCritPercentStdDev(stats: ComputedSetStats, forcedCrit: boolean, forcedDh: boolean) {
     return multiplyValues(
-        chanceMultiplierStdDev(stats.critChance, stats.critMulti),
-        chanceMultiplierStdDev(stats.dhitChance, stats.dhitMulti)
+        forcedCrit ? chanceMultiplierStdDev(1, stats.critMulti) : chanceMultiplierStdDev(stats.critChance, stats.critMulti),
```
VioletHynes, line 412: This could be expressed as e.g. `chanceMultiplierStdDev(forcedCrit ? 1 : stats.critChance, stats.critMulti)`, but I think it reads more immediately obviously this way.

xpdota:
Looks good. Would you mind adding some tests for auto-dh to validate against known results?

amakiirs (commented):
I don't have any major complaints, and nor did I see any particular issues rotationally, but I don't want to approve immediately due to my lack of detailed WAR knowledge. If no one with more experience looks at it, I can give another pass and read up more.

packages/core/src/sims/tank/war/war_sheet_sim.ts
```diff
@@ -157,6 +154,8 @@ class WarCycleProcessor extends CycleProcessor {
     }
 
     inBurst(): boolean {
+        // Six seconds after every (i.e. 0:06, 2:06, etc) burst, buffs will be up,
+        // and will remain up for twenty seconds.
         return this.currentTime % 126 < 20
```
amakiirs, line 159: With this explanation, I think it should instead be something closer to:
```typescript
const timeIntoTwoMinutes = this.currentTime % 120 // This isn't a great name but I don't have a better one offhand
return 6 <= timeIntoTwoMinutes && timeIntoTwoMinutes < 26
```
A code-cleaner but harder to understand version:
```typescript
// + 114 and -6 are equivalent mod 120, but because mod with a negative number yields a negative result,
// + 114 avoids awkwardness at low times that results from -6
return ((this.currentTime + 114) % 120) < 20
```

amakiirs (changes requested):
Correct me if I'm wrong, but I *think* the inBurst check isn't quite right.

This might (in principle anyway) be another thing to pull out into a separate function, but I'm not quite sure what that would look like. Something of the form `inTimeRangeMod(start, end, mod)`, maybe, but that's a bit awkward to understand the meaning of from the name, especially with three parameters.

packages/core/src/sims/tank/war/war_sheet_sim.ts
VioletHynes: Good catch -- fixed

amakiirs (approved):
Looks fine to me at this point, and I again don't see any obvious problems in the rotation generation.

packages/core/src/test/datamanager_tests.ts
```diff
@@ -14,90 +13,6 @@ function deq<T>(actual: T, expected: T) {
     assert.deepEqual(actual, expected);
 }
 
-describe('Old Datamanager', () => {
```
VioletHynes, line 17: I deleted these, as they weren't running until this PR, were failing, and this seems unused

PR #367, opened 2024-10-19
Normalize line endings on check-out/check-in

xiashtra:
Ref: https://git-scm.com/docs/gitattributes#_effects

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto


xpdota:
Is it possible to just make it use `LF` for everything?

xiashtra:
> Is it possible to just make it use `LF` for everything?

That's exactly what this does, on check-in (for text files only, in case the repository has binary files also).  On check-out, it will defer to the local environment, while maintaining LF on check-in.

xpdota:
Thanks!

PR #377, opened 2024-10-20
BLU: hide alt items that cannot be equipped

### LINKED ISSUE #356: Bug: Show equivalent alt items shows items that cannot be equipped due to BLU level cap
Show equivalent alt items shows items that cannot be equipped due to level for BLU (and potentially BST in the future).  This is due to BLU's max level being below the max level of other jobs.

xiashtra:
fixes #356


xpdota:
Thanks!

PR #379, opened 2024-10-21
[DRK Sim] Dump blood at end of fight, rename sim files

VioletHynes:
I renamed the sim to `drk_sheet_sim` to better match the other sims (there's also a good chance I add <100 support in the future), and remembered to make it dump blood if the fight is almost over. I also added a bunch of semi-colons.


xpdota:
Thanks!

PR #380, opened 2024-10-21
Finish polishing new toolbar UI

xpdota:
image
image

Toolbar sub-items are hidden by default, but can be opened by clicking the button. Does not auto-close since someone might want to look around while having a toolbar item open.


xpdota:
Not needed, rolled into the other PR

Issue #266, opened 2024-08-08
Enhancement: More lint rules


xpdota:
Should have:
- Semicolon at line ending
- Spacing before/after comma
- Spacing after flow control key keywords
- Anything else

PR #384, opened 2024-10-21
Add more Eslint rules

VioletHynes:
Some of these are opinionated, but enforce a consistent style. In some cases I prefer other styles but these are the most 'standard' and consistency is most important. Sorry for the big PR, but it kind of had to be for most of these rules.

The tl;dr of what this enforces:
- Indentation
- `===` must be used over `==` (as with `!==` over `!=`)
- Space after commas e.g. `[1, 2]` and not `[1,2]`
- No trailing spaces after lines
- Spaces, not tabs
- Semicolons added after every line
- Getters must always return a value
- Const should be used if the value doesn't change
- camelCase is enforced
- Variables must be used in the block they're defined in
- `isNaN` should be used over comparison to `NaN`
- A new line is added after every file
- Brace style is enforced as strousoup style

Closes https://github.com/xiv-gear-planner/gear-planner/issues/266


xpdota:
Thanks!

PR #386, opened 2024-10-24
Update contributing and readme

VioletHynes:
Mentioning ESLint, and updating some of the text around implementing a new sim to make it clearer that it can be implemented in TS if simple enough even for non-healer jobs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,8 +58,11 @@ create your `Simulation` instance.
 3. Optional: Write an extension of `SimSettings` if your sim
 has settings.
 4. Add a line in simulation.ts to call `registerSim(simSpec)`
-5. Optional: Add a GUI for your sim in `packages/frontend/src/scripts/sims`, and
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -50,19 +58,24 @@ create your `Simulation` instance.
 3. Optional: Write an extension of `SimSettings` if your sim
 has settings.
 4. Add a line in simulation.ts to call `registerSim(simSpec)`
+5. Optional: Add a GUI for your sim in `packages/frontend/src/scripts/sims`, and
```
xpdota, line 61: I feel like calling this "optional" might be underselling it a bit. If it's a cycle sim or usage count sim, which is what 99% of sims would be, then it needs a GUI, even if it's just using the default GUI. 

Might also be good to explain the different parts of the GUI.

VioletHynes: Thanks, this is a good point. Updated.

xpdota:
Thanks!

PR #387, opened 2024-10-25
Update set sim results gradiants to not be relative to the worst set

VioletHynes:
Preamble: I understand that as with any UX thing, differing opinions exist. As a result, if this is something you don't feel is an improvement, feel free to close this. However, I do believe this makes things better.

Today, gearsets are gradiented based on the difference between the worst and the best. This means that two very similar sets will show 'unfavourable red' even if they are very close. It also means that adding a new set affects the gradients of the others. See examples:
image
image-1

This approach instead uses a gradient based a logarithm so that the further away from best it is, the more it leans towards unfavourable. Anything 5% worse is always considered red. I'm not attached to this logarithm base (I think it could be even lower, but it seemed a good default).

See examples here:
image


xpdota:
Thanks!

PR #390, opened 2024-10-27
Restored PLD SKS Sim Settings Gui

Chromatophore:
Hi, I've restored the paladin sks sim's settings suite via the new ui class.

Hopefully this will pass lint - I've been having some trouble testing locally which I think I've resolved.

### DIFF
--- a/packages/frontend/src/scripts/sims/tank/pldsks_sheet_sim_ui.ts
+++ b/packages/frontend/src/scripts/sims/tank/pldsks_sheet_sim_ui.ts
@@ -1,0 +1,70 @@
+import { FieldBoundCheckBox, labeledCheckbox, labelFor } from "@xivgear/common-ui/components/util";
+import { BaseMultiCycleSimGui } from "../multicyclesim_ui";
+
+import { PldSKSSheetSettings, PldSKSSheetSimResult } from "@xivgear/core/sims/tank/pld/pldsks_sim";
+
+// Restoring the PLD SKS Sim settings UI, via the new SimGui class
+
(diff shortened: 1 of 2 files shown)


packages/frontend/src/scripts/sims/tank/pldsks_sheet_sim_ui.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        const opt1 = document.createElement("options1");
+        const space1 = document.createElement("space1");
+        const opt2 = document.createElement("options2");
+        const space2 = document.createElement("space2");
+        const opt3 = document.createElement("options3");
+        const space3 = document.createElement("space3");
+
+        behaviorDiv.appendChild(labeledCheckbox('Adjust Rotation for Skill Speed', sksCheck));
+        behaviorDiv.appendChild(labelFor("Strategy Options:", opt1));
```
xpdota, line 42: You don't really need labels to be matched up to anything that isn't a user input element. The labels are more for screen readers so that they can match up the input with the label, and so that clicking a label will activate a checkbox.

Chromatophore: I am not sure how to add label & spacers without doing it in this way, via labelfor. Are you suggesting that there is a better way to make these additional labels, and if so, can you suggest that alternative? Because, this is already pretty much my best effort.

xpdota: Should work fine, just isn't necessarily adding anything. I'll go ahead and merge, thanks!

PR #399, opened 2024-11-01
Fix ast sim drifting star and using in burst when it can't be used

jtcasper:
This logic was implemented in the rpr and monk sims to make sure that oGCDs that came up during spin time don't drift. By using it in the 2 minute burst function astro won't end up casting star 7-8s early at the end of the sim.


xpdota:
Thanks!

PR #409, opened 2024-11-06
[DRK Sim] Correct Living Shadow ability scaling

VioletHynes:
This PR corrects Living Shadow ability scaling, using the non-tank main stat scaling, removing racial bonuses, etc.

This does require require exposing some previously unexposed fields e.g. `gearStats`. We also need to pass RaceName around a little so that we can get its bonus and subtract it from the living shadow strength, but that doesn't seem so bad.

This should be extendable to apply correct scaling to other abilities that required it.

I added some tests that match up to observed values, both for Living Shadow and without it.

### DIFF
--- a/packages/core/src/sims/sim_utils.ts
+++ b/packages/core/src/sims/sim_utils.ts
@@ -94,11 +94,11 @@ export function getScalingOverrides(alternativeScalings: AlternativeScaling[], s
     const scalings = getDefaultScalings(stats);
     // Process alternative scalings for the ability. There may be multiple.
     if (alternativeScalings) {
-        if (alternativeScalings.find(scaling => scaling === "Living Shadow Strength Scaling")) {
+        if (alternativeScalings.includes("Living Shadow Strength Scaling")) {
             const livingShadowStrength = getLivingShadowStrength(stats.gearStats.strength, stats.racialStats) + strengthBuff;
             scalings.mainStatMulti = mainStatMultiLivingShadow(stats.levelStats, livingShadowStrength);
         }
-        if (alternativeScalings.find(scaling => scaling ===  "Pet Action Weapon Damage")) {
+        if (alternativeScalings.includes("Pet Action Weapon Damage")) {
             scalings.wdMulti = stats.wdMultiPetAction;
         }
     }
--- a/packages/core/src/sims/sim_utils.ts
+++ b/packages/core/src/sims/sim_utils.ts
@@ -53,9 +53,7 @@ export function abilityToDamageNew(stats: ComputedSetStats, ability: Ability, co
             dot: null,
         };
     }
-
     stats = combinedBuffEffects.modifyStats(stats);
-
     // TODO: can we avoid having all of these separate stat modifications?
     return {
         directDamage: ability.potency ? potencyToDamage(stats, ability.potency, ability as DamagingAbility, combinedBuffEffects) : null,
--- a/packages/core/src/sims/sim_utils.ts
+++ b/packages/core/src/sims/sim_utils.ts
@@ -1,10 +1,10 @@
 import {ComputedSetStats} from "@xivgear/xivmath/geartypes";
-import {Ability, Buff, CombinedBuffEffect, ComputedDamage, DamageResult, DamagingAbility} from "./sim_types";
-import {applyDhCritFull, baseDamageFull} from "@xivgear/xivmath/xivmath";
+import {Ability, AlternativeScaling, Buff, CombinedBuffEffect, ComputedDamage, DamageResult, DamagingAbility, ScalingOverrides} from "./sim_types";
+import {applyDhCritFull, baseDamageFull, getDefaultScalings, getLivingShadowStrength, mainStatMultiLivingShadow} from "@xivgear/xivmath/xivmath";
 import {multiplyFixed} from "@xivgear/xivmath/deviation";
 import {StatModification} from "@xivgear/xivmath/xivstats";
 
-function dotPotencyToDamage(stats: ComputedSetStats, potency: number, dmgAbility: DamagingAbility, combinedBuffEffects: CombinedBuffEffect): ComputedDamage {
+function dotPotencyToDamage(stats: ComputedSetStats, potency: number, dmgAbility: DamagingAbility, combinedBuffEffects: CombinedBuffEffect, scalingOverrides = getDefaultScalings(stats)): ComputedDamage {
     const modifiedStats = stats.withModifications((stats, bonuses) => {
         bonuses.critChance += combinedBuffEffects.critChanceIncrease;
         bonuses.dhitChance += combinedBuffEffects.dhitChanceIncrease;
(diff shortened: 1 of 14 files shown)


packages/xivmath/src/xivconstants.ts
```diff
@@ -121,7 +121,6 @@ const STANDARD_TANK: JobDataConst = {
     mainStat: 'strength',
     autoAttackStat: 'strength',
     irrelevantSubstats: ['spellspeed', 'piety'],
-    // traitMulti: TODO: Tank Mastery?
```
VioletHynes, line 124: Tank Mastery is already taken into account as part of this -- Tank Mastery gives you the tank power modifier:
```
        mainStatPowerMod: {
            Tank: 190,
            other: 237,
        },
```

packages/xivmath/src/xivmath.ts
```diff
@@ -149,8 +150,12 @@ export function detDmg(levelStats: LevelStats, det: number) {
  * @param jobStats Job stats for the job for which the computation is to be performed.
  * @param wd The weapon damage value.
  */
-export function wdMulti(levelStats: LevelStats, jobStats: JobData, wd: number) {
-    const mainStatJobMod = jobStats.jobStatMultipliers[jobStats.mainStat];
+export function wdMulti(levelStats: LevelStats, jobStats: JobData, wd: number, petAction: boolean = false) {
+    let mainStatJobMod = jobStats.jobStatMultipliers[jobStats.mainStat];
+    if (petAction) {
+        // All pet actions use a job mod of 100.
```
VioletHynes, line 156: This is currently only used in Living Shadow, but all pet actions use this. Damage wise, this should also apply to other things, AST Stellar Explosion, SMN abilities, Ninja Bunshin, Machinist Queen etc. I didn't include the scaling correction for those abilities as part of this PR, but it should be easy enough to do.

xpdota:
Logic seems good, but I'm wondering if it's cleaner to have the "AlternativeScaling" object contain the logic rather than just being a key? Having an entire new stat key that is only used by certain abilities on a single class is probably scope creeping RawStats.

VioletHynes:
There's not really a good way to resolve living shadow's scaling retroactively, unfortunately. Abilities don't know stats and Living Shadow needs to have its stats modified early (since it needs to have racial bonuses changed _and_ tank mastery _and_ party bonus -- there's not really a way to undo it afterwards -- it needs to be done when stats are finalized).

I did want to avoid adding it to Raw Stats, because I can definitely see the objection, but without a major change in how stats are handled, it didn't seem possible.

VioletHynes:
For what it's worth, `RawStats` isn't changed, just `ComputedSetStats` -- that seemed like the right place to update it as that's where stuff like party bonuses are applied. I am all ears if you can think of an alternate approach that works, but adding this as an output to a set's stats seemed like the only option.

VioletHynes:
I refactored Alternative Scalings a little to make it easier for other jobs to adopt. Makes sense to do it on a scaling-by-scaling basis to minimize code duplication.

packages/xivmath/src/xivstats.ts
```diff
@@ -406,6 +426,20 @@ function finalizeStatsInt(
...
+    if (racialBonus) {
+        // Remove the 'base' race bonus of 20.
+        livingShadowRaceBonus -= 20;
+        const strengthBonus = racialBonus['strength'];
+        if (strengthBonus) {
+            livingShadowRaceBonus -= strengthBonus;
+        }
+    }
+    // Living Shadow does not take into account party bonus, and uses +2 racial
+    combinedStats.livingShadowStrength = combinedStats['strength'] + livingShadowRaceBonus;
```
xpdota, line 441: This section here is the only part that is actually unique to living shadow, right? So could it apply some kind of stat modification externally, like:

```ts
    stats.withModifications((stats, bonuses) => {
        let livingShadowRaceBonus = 2;
        const racialBonus = stats.racialBonus; // this isn't currently on stats but it could be added
        if (racialBonus) {
            // Remove the 'base' race bonus of 20.
            livingShadowRaceBonus -= 20;
            const strengthBonus = racialBonus['strength'];
            if (strengthBonus) {
                livingShadowRaceBonus -= strengthBonus;
            }
        }
        bonuses['strength'] -= livingShadowRaceBonus;
    });
```

And then it could just be applied in abilityToDamageNew alongside applying buff effects, which also gives you control over the order of operations there.

VioletHynes: The primary one is the main stat multiplier (that uses this strength as an input) -- Living Shadow uses a separate formula where the 'other' job scaling is used rather than Tank Mastery, i.e. the new function `mainStatMultiLivingShadow`.

Though, looking at this, you're right that it looks like I can do this without adding a special Living Shadow stat to every stat. I'll be busy until Sunday but I think I should be able to get a version two of this PR that only adds racial stats to the stats.

Thanks for the feedback! I did feel like my implementation was a little ugly so I'm happy it looks like there's a better way.

VioletHynes: Hmm, wait. On second thought, I still need to remove the party bonus from Living Shadow's strength, so I'll need to pass the party bonus amount into the stats too, and I have to do that before other modifiers are applied (i.e. the suggestion of computing the strength post-hoc is severely complicated by the fact most other bonuses happen _after_ party bonus). I'm less convinced it'll be able to be completely non-ugly, but maybe I can make it better.

VioletHynes:
Okay, I made it not terrible. 

I had to expose a couple extra things but they're not just useful for Living Shadow, but other abilities (like Bunshin). It's still missing one thing, which is that Living Shadow's stats aren't modified by potions, but that should be fixable, I just want to make sure this approach is agreeable before I go further. Potions are a little trickier to solve given where we apply buffs but I'm pretty sure I can find a way to make it work.

packages/core/src/sims/sim_utils.ts
```diff
@@ -80,6 +87,25 @@ export function noBuffEffects(): CombinedBuffEffect {
...
 }
 
+/**
+ * Translates the given alternate scalings into the specific numerical scaling overrides.
+ */
+export function getScalingOverrides(alternativeScalings: AlternativeScaling[], stats: ComputedSetStats, strengthBuff: number = 0): ScalingOverrides {
+    const scalings = getDefaultScalings(stats);
+    // Process alternative scalings for the ability. There may be multiple.
+    if (alternativeScalings) {
+        if (alternativeScalings.find(scaling => scaling === "Living Shadow Strength Scaling")) {
```
xpdota, line 97: This can be `alternativeScalings.includes("Living Shadow Strength Scaling")`

VioletHynes:
Okay, I added support for pots too, and I think this should be good to go, pending feedback of course. Let me know if you'd like me to adjust the approach in any part. Appreciate the review and suggestions so far!

packages/core/src/sims/sim_utils.ts
VioletHynes: Thanks!

xpdota:
Awesome, thanks!

PR #410, opened 2024-11-08
Added a basic healer MP economy calculator

Mayceia:
I haven't set it to show by default since it's pretty basic but it should be useful.

### DIFF
--- a/packages/core/src/sims/healer/healer_mp.ts
+++ b/packages/core/src/sims/healer/healer_mp.ts
@@ -14,7 +14,7 @@ const fillerMP = 400;
 const drawCooldown = 60;
 const drawMP = 2000;
 
-/*const aetherflowCooldown = 60;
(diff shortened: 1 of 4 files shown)


packages/core/src/sims/healer/healer_mp.ts
```diff
@@ -0,0 +1,95 @@
...
+    loadSavedSimInstance(exported: MPSettings) {
+        return new MPPerMinute();
+    },
+    makeNewSimInstance(): MPPerMinute {
+        return new MPPerMinute();
+    },
+    stub: "mp-sim",
+    description: "Mp economy",
+    isDefaultSim: false,
+    supportedJobs: ['AST', 'SCH', 'SGE', 'WHM'],
```
xpdota, line 51: This seems like it is AST-specific?

xpdota:
Thanks!

PR #414, opened 2024-11-11
[AST Sim] Correct Earthly Star scaling

VioletHynes:
Before:
image

After:
image

### DIFF
--- a/packages/core/src/sims/healer/ast_sheet_sim.ts
+++ b/packages/core/src/sims/healer/ast_sheet_sim.ts
@@ -53,6 +53,7 @@ const star: AstOgcdAbility = {
     name: "Earthly Star",
     potency: 310,
     attackType: "Ability",
+    alternativeScalings: ["Pet Action Weapon Damage"],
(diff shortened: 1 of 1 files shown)


Mayceia:
Going off the pet scaling sheet diastarvy made a few months ago this might actually make it less accurate
https://docs.google.com/spreadsheets/d/1Yt7Px7VHuKG1eJR9CRKs3RpvcR5IZKAAA3xjekvv0LY/edit?gid=0#gid=0
Earthly Star's tooltip potency is a lie that ends up pretty close to character potency, so if we're fixing it we probably need to also remove maim and mend scaling from it and put in the real potency in the spell data

VioletHynes:
Hmm, right -- I'd missed that complication. I'll close this for now, but I might re-attempt at a future point this based on known values if I'm feeling more ambitious as I don't want to leave this in a half-finished state. Though, feel free to continue this in my stead -- the framework should be there for alternate scalings now and you probably know more about Earthly Star's intricacies than me.

PR #420, opened 2024-11-14
Fix sim caching when settings change

### LINKED ISSUE #389: Bug: Cached results persist when updating Sim Settings
If you modify a sim's settings, it appears the cached results of the sim will persist, with no clear method to re run it. Since all sims have Party Buff options, all sims are probably impacted.

The 2 actions I've found that will re run sims are
a) Changing gear
b) Refreshing the page

All exported sheets, such as this one for Samurai:
https://xivgear.app/?page=sl%7C78944845-857b-4a2f-bc04-576f4c6d9dec
_Do_ allow you to access the sim settings, so, you can turn on party buffs, change potion usage etc, but, since it's an exported set, you can neither change the gear set, or refresh the page, as doing so naturally restores the settings.

Additionally, where you have an editable set, it appears that if you change the sim settings for "Use Auto Attacks" and the "Total Time", for any sims using the newish CycleSettings class, your choices for these seem to be overridden and restored to those defined as default in the noted class when the sim loads, so, eg, on refresh.

Here are some repro steps:

1) Make a new sheet for Dark Knight (DRK), and give it some gear.
2) Open the sim's sheet by clicking the dps number, observe the settings to be active (Drk first action = TBN)
3) Access sim settings, turn off potion/use tbn, turn off auto attacks, set fight time to 7 seconds, add a ninja.
4) Observe number hasn't changed, open the sheet, observe TBN, pot, more than 7 seconds of fighting, no ninja buff.
5) Refresh the page
6) Observe the settings page again; Sim Specific & Party Comp should match above, however, Cycle Settings have been restored (DRK's current default values are 510 & Use Auto-Attacks)
7) Observe the table, no TBN or potion in the case of DRK, but, normal length encounter & autos.

I believe expected behavior is that
step 4 should show you an updated dps number, with a 7 second table with no TBN,
and
step 6 & 7 would retain your choice of no autos and a 7 second fight.

xpdota:
Fixes #389

### DIFF
--- a/packages/core/src/sims/healer/healer_mp.ts
+++ b/packages/core/src/sims/healer/healer_mp.ts
@@ -1,7 +1,7 @@
-import { CharacterGearSet } from "@xivgear/core/gear";
-import { SimResult, SimSettings, SimSpec, Simulation } from "@xivgear/core/sims/sim_types";
-import { NORMAL_GCD } from "@xivgear/xivmath/xivconstants";
-import { EmptyObject } from "../../util/types";
+import {CharacterGearSet} from "@xivgear/core/gear";
+import {SimResult, SimSettings, SimSpec, Simulation} from "@xivgear/core/sims/sim_types";
+import {NORMAL_GCD} from "@xivgear/xivmath/xivconstants";
+import {EmptyObject} from "../../util/types";
 
 const maxMP = 10000;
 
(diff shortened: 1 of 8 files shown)


packages/core/src/sims/healer/healer_mp.ts
```diff
@@ -1,7 +1,7 @@
-import { CharacterGearSet } from "@xivgear/core/gear";
-import { SimResult, SimSettings, SimSpec, Simulation } from "@xivgear/core/sims/sim_types";
-import { NORMAL_GCD } from "@xivgear/xivmath/xivconstants";
-import { EmptyObject } from "../../util/types";
```
xiashtra, line 4: Why are we still getting whitespace changes?  This should have all been normalized by the linter update in #384

xpdota: I don't know, it doesn't seem to have a rule for that.

PR #426, opened 2024-11-15
Default meld solver to current GCD, small UX tweaks

VioletHynes:
Meld solver now defaults to target GCD being toggled as true, and displays the number to two decimal places (i.e. `2.50` instead of `2.5`.

After:
image

### DIFF
--- a/packages/common-ui/src/components/util.ts
+++ b/packages/common-ui/src/components/util.ts
@@ -172,6 +172,14 @@ export interface FbctArgs<ObjType, FieldType> {
     inputMode?: string;
 }
 
+export interface FieldBoundFloatFieldFbctArgs<ObjType, FieldType> extends FbctArgs<ObjType, FieldType> {
+    /**
+     * Optionally, how many decimal places to fix the float to. You can use 2 to make GCD speeds (i.e. 2.50)
+     * show up as 2.50 instead of 2.5.
+     */
+    fixDecimals?: number;
+}
+
 export class FieldBoundConvertingTextField<ObjType, FieldType> extends HTMLInputElement {
 
     reloadValue: () => void;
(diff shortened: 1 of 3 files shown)


packages/frontend/src/scripts/components/materia.ts
```diff
@@ -415,7 +415,7 @@ export class MateriaPriorityPicker extends HTMLElement {
                     ctx.failValidation("Cannot be greater than " + MAX_GCD);
                 }
             }],
-        });
+        }, true);
```
VioletHynes, line 418: Only marginally related to the meld solver, but it also displays a GCD speed, so I applied the `.toFixed(2)` to this, too

packages/common-ui/src/components/util.ts
```diff
@@ -364,7 +364,7 @@ export class FieldBoundIntField<ObjType> extends FieldBoundConvertingTextField<O
 }
 
 export class FieldBoundFloatField<ObjType> extends FieldBoundConvertingTextField<ObjType, number> {
-    constructor(obj: ObjType, field: { [K in keyof ObjType]: ObjType[K] extends number ? K : never }[keyof ObjType], extraArgs: FbctArgs<ObjType, number> = {}) {
+    constructor(obj: ObjType, field: { [K in keyof ObjType]: ObjType[K] extends number ? K : never }[keyof ObjType], extraArgs: FbctArgs<ObjType, number> = {}, displayTwoDecimalPlaces = false) {
```
VioletHynes, line 367: I thought of making the new argument a number to string function, but it felt like a premature optimization, and showing GCD speed seems like by far the primary use case for any nonstandard display.

xpdota: I think it would make sense to extend `FbctArgs` and add a number of decimal places arg. It could also be extended to change the pattern used for the field to enforce a particular decimal count, like what the meld solver UI is currently doing but without needing to do it in two places.

VioletHynes: This is a much better idea than what I had -- fixed. Let me know if this looks good

xpdota:
Thanks!

PR #427, opened 2024-11-17
Fix Meld Solver default target GCD

Junie-774:


packages/frontend/src/scripts/components/meld_solver_modal.ts
```diff
@@ -188,25 +188,19 @@ class MeldSolverSettingsMenu extends HTMLDivElement {
...
-            const haste = set.computedStats.haste(override[0].attackType) + (override[0].haste ?? 0);
-            switch (override[0].basis) {
-                case "sks":
-                    gcd = set.computedStats.gcdPhys(2.5, haste);
-                    break;
-                case "sps":
-                    gcd = set.computedStats.gcdMag(2.5, haste);
-                    break;
-            }
+            haste += (override[0].haste ?? 0);
```
xpdota, line 194: I'm worried that this is completely ignoring `set.computedStats.haste` now.

Junie-774: `haste` is initialized with `set.computedStats.haste`

xpdota:
Thanks!

PR #429, opened 2024-11-19
[GNB Sim] Gunbreaker Sim

VioletHynes:
A Gunbreaker Sim. It's not perfect, but it performs everything admirably and gets all of its important abilities in No Mercy.

For weird speeds where things don't align perfectly (e.g. 2.45), it opts to clip (if the clip is caused by GCD speed) to clip to keep the GCDs from drifting. This seems to be better than any alternative I could find.

image

image


xpdota:
Idr, were you able to fix the timing issue Krom pointed out?

VioletHynes:
> Idr, were you able to fix the timing issue Krom pointed out?

Yeah -- at some speeds it just micro-clips, but we're pretty sure that's how it is in reality, too (logs aren't accurate enough time wise). It's addressed by the code around this bit, which micro-clips if it's caused by a speed-based microclip:

>const shouldUseGCD = relativeCooldown === 0  || relativeCooldown <= (gcdSpeed / 20);

All of the numbers look as expected now (and stuff doesn't drift and is in No Mercy as expected).

xpdota:
Thanks!

Issue #431, opened 2024-11-21
Enhancement: Typed Haste


raineaeternal:
While I was looking into Riddle of Wind for the Monk sim, I found that `combinedEffects.haste` and potentially `stats.bonusHaste("Auto-Attack)` does not accurately implement different types of haste. The problem this causes is that Monk damage will be inflated on the sim as a result, these should be checked for their haste type.

AkhMorning outlines Greased Lightning to be a Type Z buff, while Riddle of Wind *appears* to be a Type Y buff. Monk is so far the only job to interact with this kind of buff shenanigans and more testing and research would be needed to accurately prove this. from my limited testing, however, it seems to be that the FFLogs estimate only produces a 0.03s disparity between RoW and non-RoW Auto Attack Delays. (This may be due to how the auto attack functions in game.)

This work would also benefit other jobs, like Black Mage and other jobs with conditional haste buffs. (ex. Leylines)

If you need more data, or just have a chat about this, do feel free to @ me in The Balance, or dm me; `raineaeternal`

PR #433, opened 2024-11-21
Scaffold for haste types so that monk has correct auto delay calculations

jtcasper:
#431

I've been investigating the type of Riddle of Wind's haste buff, and Aya Liz confirms that RoW should have an auto delay of 1.024. The only way for that to be true is for RoW and GL4 to have different haste types as defined by Allagan Studies. This bug currently only affects monk, as no other job has two sources of haste.

I've left backwards compatibility in for any `HasteBonus`es or traits that aren't typed, but with approval I can go through and modify these.

### DIFF
--- a/packages/core/src/sims/cycle_sim.ts
+++ b/packages/core/src/sims/cycle_sim.ts
@@ -1202,8 +1202,17 @@ export class CycleProcessor {
             snapshotTimeFromStart: 0,
             lockTime: 0,
         });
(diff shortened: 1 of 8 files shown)


xpdota:
I feel like if the different haste types are independent, it would make more sense for it take one of two approaches:

1. Have separate bonusHaste lists (bonusHaste, bonusHasteY, bonusHasteZ), or
2. Have the `apply` function return (number, type) instead of having the type be external.

PR #435, opened 2024-11-29
[PLD Sim] Add PLD cycle sim, make it the default PLD sim, normalize PLD stuff

VioletHynes:
This PR does a few things:
- Renames some of the PLD files to have the same naming conventions as other jobs
- Adds a new default PLD sim, `pld-sheet-sim`, a cycle sim for Paladin
- Made the usage sim for PLD not default, and added a description to it (hence, the new sim is the only default)
- I _could_ have re-tooled the experimental tactics SKS sim for this, but I thought it might be an overuse of that sim, and PLD logic is extremely simple so I thought it made more sense to have the default/'main' sim for comparing gearsets as a new one without the extra experimental stuff so the code and settings could be cleaner
- This means that all tanks will have a default cycle sim with a 8m 30s killtime, which also allows for really easy comparisons between all of the tanks, which I'm personally very excited for

The new sim is optimal as far as I can tell, and performs as good or better than the existing sim at most speeds in rDPS (this is with pots turned off to match the existing sim, though the new sim does use pots correctly, and has a toggle)
image

With buffs taken into account, the new sim wins out in aDPS in every situation I can find (this would be more stark with pots, but the other sim doesn't support pots):
image

With pots:
image

It works, and is good:
image

I do realize some of the decisions here are opinionated, and I'm open to feedback, but I think this makes things better overall


packages/core/src/sims/tank/pld/pld_sheet_sim_no_sks.ts
VioletHynes, line 1: There didn't seem much reason to keep this cycle sim that doesn't take into account SkS at all, now that there's a replacement

packages/frontend/src/scripts/sims/tank/pld_sheet_sim_ui.ts
```diff
@@ -0,0 +1,75 @@
...
+import {BaseMultiCycleSimGui} from "../multicyclesim_ui";
+import {AbilitiesUsedTable} from "../components/ability_used_table";
+import {CycleSimResult, DisplayRecordFinalized, isFinalizedAbilityUse} from "@xivgear/core/sims/cycle_sim";
+import {CustomColumnSpec} from "../../tables";
+import {PreDmgUsedAbility} from "@xivgear/core/sims/sim_types";
+import {PldExtraData} from "@xivgear/core/sims/tank/pld/pld_types";
+import {PldSettings, PldSimResult} from "@xivgear/core/sims/tank/pld/pld_sheet_sim";
+
+export class PldSimGui extends BaseMultiCycleSimGui<PldSimResult, PldSettings> {
+    static generateResultColumns(result: CycleSimResult): CustomColumnSpec<DisplayRecordFinalized, unknown, unknown>[] {
```
VioletHynes, line 11: I opted to only show FoF duration, and not e.g. Requiescat stacks, but PLD doesn't have a lot to show

packages/frontend/src/scripts/sims/tank/pld_sheet_sim_ui.ts
```diff
@@ -0,0 +1,75 @@
...
+                        barOuter.style.borderRadius = '20px';
+                        barOuter.style.background = '#00000033';
+                        barOuter.style.width = '120px';
+                        barOuter.style.height = 'calc(100% - 3px)';
+                        barOuter.style.display = 'inline-block';
+                        barOuter.style.overflow = 'hidden';
+                        barOuter.style.border = '1px solid black';
+
+                        const barInner = document.createElement('div');
+                        barInner.style.backgroundColor = '#B14FAE';
```
VioletHynes, line 41: I took this colour from the FoF icon :)

packages/core/src/sims/tank/pld/pldsks_sim.ts
```diff
@@ -358,7 +358,7 @@ export class PldSKSSheetSim extends BaseMultiCycleSim<PldSKSSheetSimResult, PldS
     };
 
     spec = pldSKSSheetSpec;
-    shortName = "pld-sheet-sim";
+    shortName = "pldsks-sheet-sim";
```
VioletHynes, line 361: This matches the stub of the sim.

VioletHynes:
Updated with an optimization for 2.44 and 2.42.
Before (without pots in both cases for a fairer comparison)
image

After
image

xpdota:
Thanks!

Issue #238, opened 2024-07-28
Feature: Pictomancer Simulator


avantearpeggio:
Add support for Pictomancer to xiv-gear-planner.

avantearpeggio:
#237 - To add party buffs.

avantearpeggio:
Progress check-in:
* Initial picto sim is ready and seems to work properly and all buffs apply properly

Bugs:
* Starry unintentionally drifts, likely need to do something for it to pull it back a GCD.

Next steps:
* Fix bug in starry drift OR swap to a hardcoded rotation for 2.50
* Add gauges for Palette (steal from NIN most likely)

Future:
* Add separate 2.48 rotation like SAM if I can't fix the drift/end-time to make 2.48 and 2.50 more directly comparable.
* Add gauge for white paint and monochrome tracker (not really that big of a deal considering picto will naturally overcap on white paint)

Preview: https://github.com/avantearpeggio/gear-planner

PR #436, opened 2024-12-01
[PCT Sim] Add initial Picto consts, types, actions, buffs, and a primitive gauge.

avantearpeggio:
Add the skeleton for a Pictomancer simulator for https://github.com/xiv-gear-planner/gear-planner/issues/238.

Adds in:

* Actions
  * All ST Spells
  * All relevant oGCDs
* Initial logic for applying and removing buffs
* Skeleton Gauge with no frills


avantearpeggio:
Note for myself tomorrow: these files should be moved to core, they ended up in UI (maybe an old location that I never updated in the months since I started down this rabbit hole)

avantearpeggio:
Okay, moved files as-is to core and resquashed.

PR #439, opened 2024-12-05
[WAR Sim] Add support for 70-90 WAR

VioletHynes:
I tested all of the levels, and it Seems Good. It's not doing anything it shouldn't, numbers look accurate, etc.

UWU:
image

TEA:
image

DSR: 
image


xpdota:
Thanks!

PR #441, opened 2024-12-06
[PLD Sim] 70, 80, 90 support for PLD

VioletHynes:
Everything seems to work, it's not using incorrect abilities, and the numbers look correct and expected. (and I didn't break anything for level 100 PLD, for any sim)

UWU:
image

TEA:
image

DSR:
image


xpdota:
Thanks!

PR #446, opened 2024-12-17
Settings cog is clickable

hintxiv:


xpdota:
Does this affect the whole cell, or just the gear? The entire cell is clickable, the gear is just there to provide a visual hint that it's clickable.

PR #447, opened 2024-12-17
Sim result is clickable

hintxiv:


xpdota:
Shouldn't this affect the entire <td>, since there's no guarantee that the <sim-result-display> will take up the entire <td> space.

PR #448, opened 2024-12-17
Buttons are clickable

hintxiv:


xpdota:
I think this might be overly-broad. `.like-a-button` is used for some things that look like buttons but are not interactive.

PR #449, opened 2024-12-17
Stat tier buttons are clickable

hintxiv:


xpdota:
Thanks.

Note to self: make the primary stat tier display not clickable at all, since it doesn't have a dropdown.

PR #450, opened 2024-12-17
Rows in editable tables are clickable

hintxiv:


xpdota:
This is too broad, it seems to also affect the header row. I'd assume that the `not(:first-child)` would take care of that, but the "Header" of each table with the slot name and chevron are actually the first child.

I think it makes more sense to do the rows based on whether or not they contain a `<td> child, as the "header" and the actual table header only contain <th>s.`

xpdota:
Fixed in #704

PR #453, opened 2024-12-17
Accordion carets are clickable

hintxiv:


xpdota:
This is too narrow as it is actually the entire horizontal bar that is clickable.

xpdota:
Fixed in #704

PR #454, opened 2024-12-17
Add hover styles to main menu SVGs

hintxiv:


xpdota:
I can't seem to notice a difference with this

PR #458, opened 2024-12-20
Make role bonus for new sheets 5 default (or 1 for BLU)

VioletHynes:
Makes the default role bonus 5 for non-BLU jobs and 1 for BLU, to better be a (in my opinion) better and more expected default.


xpdota:
Thanks!

PR #470, opened 2025-01-25
Lvl sync display

xpdota:
Ilvl sync only (unchanged):
image

Both lvl and ilvl sync: 
image

Lvl sync only, no explicit ilvl sync (displays implicit ilvl sync for the weapon slot):
image


xiashtra:
This looks fine to me; alternatively, we could just always show the level, even if not synced, for maximum end-user clarity.

PR #472, opened 2025-01-31
Draft: Handle iaijutsu cast times not scaling with haste

jtcasper:
As seen/discussed in Discord
image

### DIFF
--- a/packages/core/src/sims/sim_types.ts
+++ b/packages/core/src/sims/sim_types.ts
@@ -324,6 +324,12 @@ export type BaseAbility = Readonly<{
      */
     fixedGcd?: boolean,
     /**
+     * If the ability has a fixed cast _but_ a scaling recast, set to true.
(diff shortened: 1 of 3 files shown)


packages/core/src/sims/sim_types.ts
```diff
@@ -323,6 +323,12 @@ export type BaseAbility = Readonly<{
      * set to true.
      */
     fixedGcd?: boolean,
+    /**
+     * If the ability has a fixed cast _but_ a scaling recast, set to true.
+     * The only current occurrence of this in game is the Iaijutsu and Ogi Namikiri weaponskills after
+     * the level 74 talent.
+     */
+    fixedCast?: boolean,
```
xpdota, line 331: Can this be changed to something like `fixedCastOnly`?

Also I can fix it later, but if this is supposed to be mutually exclusive with `fixedGcd`, the type definition should enforce that.

PR #476, opened 2025-02-13
Set size of the worker pool based on number of logical processors available

xiashtra:
Improves meld solving speed on systems with > 4 cores (previous fixed worker pool size).

As an aside: Mocha is already pulling in [workerpool](https://www.npmjs.com/package/workerpool) as a dependency; it might have made more sense to use that instead of rolling our own implementation.

### DIFF
--- a/packages/frontend/src/scripts/workers/worker_pool.ts
+++ b/packages/frontend/src/scripts/workers/worker_pool.ts
@@ -231,6 +231,8 @@ export class WorkerPool {
     }
 }
 
-// TODO: pool size
-// Going too high seemed to slow it down even on a system with more than enough cores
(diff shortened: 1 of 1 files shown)


xpdota:
LGTM.

Yeah, it might make sense to just look at workerpool, mostly because the current implementation keeps a fixed amount of workers around even if they're idle. They all use a baseline amount of memory even if they're doing nothing.

xpdota:
I looked into workerpool a bit. One of the disadvantages of it for the time being is that it doesn't have as much type checking around your TS types for work requests. The other issue is that I'm not sure if it supports an explicit "initialization" step. It seems that it expects the workers to be stateless.

PR #482, opened 2025-02-27
Fix project structure, enable TS strict mode in most places

xpdota:
- Enables Typescript strict mode for most of the modules
- Introduces a couple more modules
- Fixes things to make it pass strict validation
- Refactor some types to make them pass strict validation


packages/core/src/sims/common/potion.ts
```diff
@@ -37,12 +37,14 @@ function makePotion(name: string, stat: RawStatKey, itemId: number, bonus: numbe
...
+
+export function makeGemdraught(stat: Mainstat, grade: GemdraughtGrade): Readonly<OgcdAbility> {
     const statToPotItemId = {
         mind: 44161,
         strength: 44157,
         dexterity: 44158,
         intelligence: 44160,
+        vitality: 44159,
     };
     const gradeToStatCap = [351, 392];
```
xiashtra, line 49: Vitality pots have a different stat cap

PR #487, opened 2025-03-20
Always show level in sync info it item level is synced

VioletHynes:
This will always show level in the sync info if the item level is synced. Mostly, this helps prevents silly mistakes like forgetting to sync level and makes everything a lot more explicit, so I think this is generally a positive improvement. I think future work would include making 'strange' syncs' warn users, but I think this is a positive even in that world.

image

(Works in view mode too)

`tsconfig-paths-webpack-plugin` seems to be required to build, and it wasn't present in a package.json (npm install didn't install it) so I added it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,15 +8,15 @@
       "name": "gearplan",
       "version": "1.0.0",
       "workspaces": [
+        "packages/util",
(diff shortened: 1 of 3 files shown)


xpdota:
Thanks!

PR #490, opened 2025-03-24
Add monk 7.2 potency bumps

jtcasper:
Wind's Reply
	Potency has been increased from 900 to 1,040.
Reduction in potency after the first target has been changed from 50% to 40%.
Fire's Reply
		Potency has been increased from 1,200 to 1,400.
Reduction in potency after the first target has been changed from 50% to 40%.


xpdota:
Thanks!

PR #502, opened 2025-03-28
Add new default-on option to ignore Gunbreaker Gnashing Fang microclips

VioletHynes:
The current sim will microclip (as expected) for Gunbreaker, i.e. intentionally clip its GCD due to the weird scaling of Gnashing Fang. This isn't 'incorrect' per se, because gaining more of a GCD is better in a vacuum, but the killtime dependency means that, practically, comparing sets at different speeds means the sim results show more value in SkS melds than it probably should.

This might feel like a bit of an ugly solution, but to me it feels like the right one (I first tried the opposite approach, being pessimistic and making all microclips as bad as possible). Due to the fact that FPS/ping/etc exists, this more closely matches the reality of playing the job (i.e. 2.50's fps/ping tax is just raw clipping whereas 2.45s will eat into the microclip and not be as bad as the sim shows), and values things a lot more sensibly for the purposes of comparing gearsets, which is what the sim is for, first and foremost. 

Example (note that the sets are higher than prod because of grade 3 gemdraught):
The first copy of 2.45 has a crit meld removed.
The second copy has a SkS meld added, it has slightly higher DPS (due to auto attack scaling) but not as much as it would have otherwise gained, and not more than a normal meld.
image

### DIFF
--- a/packages/core/src/sims/tank/gnb/gnb_sheet_sim.ts
+++ b/packages/core/src/sims/tank/gnb/gnb_sheet_sim.ts
@@ -20,6 +20,10 @@ export interface GnbSettings extends SimSettings {
     usePotion: boolean;
     // the length of the fight in seconds
     fightTime: number;
+    // If set, will make the rotation 'cheat' and assume that Gnashing
+    // Fang and Double Down align perfectly with the GCD.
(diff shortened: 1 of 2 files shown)


xpdota:
Thanks!

PR #534, opened 2025-04-13
Make Meld Solver button first-class, rename materia fill/solve to materia fill/lock

VioletHynes:
There are a few problems with including the meld solver in the fill/lock button in my eyes:
- Users get confused that the fill priority affects the meld solver
- This is one of the most awesome features of the site, and hiding it in a sub-menu can cause people to not know about it

This PR makes the meld solver button first-class, and moves it out of the fill/lock button. Making it more obviously disconnected from the 'fill' capabilities, as well as being more obvious to users.

image
image
image


xpdota:
Thanks!

PR #537, opened 2025-04-15
Fix Nascent Chaos status ID

VioletHynes:
50267 isn't a real ID, correcting it to 1897 (what it should be). See also: https://www.garlandtools.org/db/#status/1897


xpdota:
Thanks!

PR #544, opened 2025-04-17
[GNB Sim] Add hypervelocity no mercy optimization for 2.50

VioletHynes:
image

Specifically does this for 2.50, whereas previously it'd use Brutal Shell. Minor gain to 2.50 DPS, other speeds untouched.


xpdota:
Thanks!

PR #548, opened 2025-04-18
Change food item level default to 740

xiashtra:
Also changes to use the same default food item level at each level sync bracket.


xpdota:
Thanks!

PR #551, opened 2025-04-20
Update index page message to be a little less outdated

VioletHynes:
Just wanted to make this a little more up to date and accurate.


xpdota:
Thanks!

PR #565, opened 2025-04-23
Fix blitz window entry for "-3" opos

jtcasper:
While implementing the triple lunar rotation I noticed that the 8 minute burst window entry was incorrect in all rotations due to drift / GCD alignment. This caused all rotations to desync brotherhood and rof double weaves while entering the 8 minute window, because the blitz entry logic was not considering windows that would allow the first GCD under buffs to be a blitz.


xpdota:
Thanks!

PR #580, opened 2025-05-01
Make sim info consistent in both view and edit mode by moving it to the left in edit mode

VioletStardustXIV:
As per title. Makes sense for them to be consistent and it is the most salient piece of info in both.


xpdota:
Thanks!

Issue #582, opened 2025-05-03
Solving does not preserve locked/unlocked slots


Junie-774:
**Describe the bug**
When you solve melds for a set which has some slots locked empty, the resulting set respects the locked slots, but they are no longer locked afterwards. (i.e. locked slots are still empty/whatever they were before, but now unlocked

**Expected behavior**
Expected behavior would be that all slots that were locked before melding are still locked after melding

**Browser**
Chrome

xpdota:
Fixed in #583 

(I think GH expects the syntax to be exactly `Fixes #582` rather than `Fix for #582` for the auto-close to work)

PR #583, opened 2025-05-03
Fix solver not preserving locked slots

Junie-774:
Fix for https://github.com/xiv-gear-planner/gear-planner/issues/582


xpdota:
Thanks!

PR #597, opened 2025-05-12
Initial BLM sim implementation

clo-yunhee:
Only level 100 for now but planning to add lower levels later. Also need to clean up the code.
The way it weaves oGCDs is quite bruteforcey but it works since it's just a dummy rotation.

The results themselves look pretty accurate* to the "7.X BLM Relative Damage Calc" sheet for the few gearsets that I've tried which is promising. *or at least in the same ballpark as them


packages/core/src/sims/caster/blm/blm_gauge.ts
```diff
@@ -0,0 +1,184 @@
+import {BlmGaugeState} from "./blm_types";
+
+const damageUpModifiers = [1.4, 1.6, 1.8];
+const damageDownModifiers = [0.9, 0.8, 0.7];
+
+export class BlmGauge {
+
+    private _aspect: number = 0;
```
VioletStardustXIV, line 8: This is a bit of a nit but I think it'd be great if there were constants throughout for especially F3 and B3, e.g.
`export const BlizzardThreeAspect = -3;` or whatever

packages/core/src/sims/caster/blm/blm_sheet_sim.ts
```diff
@@ -0,0 +1,739 @@
...
+        else if (cp.stats.level === 90) {
+            //this.useLevel90Opener(cp, prepullTBN);
+        }
+        else {
+            //this.useLevel80OrBelowOpener(cp, prepullTBN);
+        }
+    }
+
+    getRotationsToSimulate(set: CharacterGearSet): Rotation<BlmCycleProcessor>[] {
+        const gcd = set.results.computedStats.gcdPhys(2.5);
```
VioletStardustXIV, line 722: This should be `gcdMag`

VioletStardustXIV (commented):
This looks awesome! Have a few small clean-up comments and while I'm not the biggest BLM expert, the logic looks at the very least at a basic level good to me. 

This seems very well-made and thorough though. Awesome job so far!

(I'm not an approver or maintainer, that'd be Wynn, but I've done a lot of sims so I thought my review would be valuable regardless)

packages/core/src/sims/caster/blm/blm_gauge.ts
clo-yunhee: Yeah I'll probably refactor this into a proper type at some point but it felt way too verbose to prototype with so I just stuck with this +3/-3 thing for now

PR #611, opened 2025-05-15
[DRK] Correct Living Shadow scaling, make the code better

VioletStardustXIV:
The old one was 'correct' by accident at level 100 because 440 * 1.05 = 462 which is suspiciously similar to +22. tl;dr:
- Living Shadow was 2 strength off at level 100
- It was incorrect by I can't be bothered to figure it out at 90 and 80

Correct value is ⌊440 * 1.00⌋ + 2, and correct base for player is ⌊440 * 1.05⌋ + RacialOffset

I also feel like the code was messy/bad so I cleaned it up a bit. No reason to re-get racial stats when they're right there. This also should help us better split up gear stats and non-gear stats if we ever need to in the future, also.

### DIFF
--- a/packages/core/src/gear.ts
+++ b/packages/core/src/gear.ts
@@ -40,7 +40,7 @@ import {
     XivCombatItem
 } from "@xivgear/xivmath/geartypes";
 import {Inactivitytimer} from "@xivgear/util/inactivitytimer";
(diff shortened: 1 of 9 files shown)


xpdota:
Thanks!

PR #612, opened 2025-05-15
Add level 80 and 90 monk support by updating potencies and openers

jtcasper:
This copies the only differences from the new abilities monk gains from 91-100 to the basic GCDs that they are upgrades for.

I checked that this behaves identical to live for level 100 using the MNK bis sheet and the opener looks good at 90, however without published resources for 90 openers it's a best effort.


jtcasper:
Leaving this in draft until #616 and #614 land (or at least #614 with bunk status ids is approved)

PR #676, opened 2025-06-20
Don't add ' copy' to the name of newly imported sheets

VioletStardustXIV:
Live:
image

Local:
image

Still adds 'copy' for exported sheets, or existing saved sheets
Exported:
image

Existing:
image


xpdota:
Thanks!

PR #678, opened 2025-06-22
Always show level sync info on small screens

VioletStardustXIV:
I also made the decision to move it to the front in edit mode too, so that it wouldn't move around.
Before:
image
image

After:
image
image


xpdota:
Thanks!

PR #690, opened 2025-07-07
Replace | and %7C with _ in xivgear URLs

VioletStardustXIV:
This commit replaces `|`/`%7C` in xivgear URLs with `_`. 

It really affects the readability of e.g. <https://xivgear.app/?page=bis%7Cdrk%7Cultimate%7Cucob#> and there have been issues in the past with knowing where the GUID starts ends for e.g. <https://xivgear.app/?page=sl%7C9d117608-acc4-415f-97fd-6c4dddbd790a> because the '7C' looks like part of it.

Advantages:
- It doesn't ever get URL encoded
- It's not part of a GUID, like `-` or alphanumeric characters
- I should make gaps in words clear, unlike e.g. `.` (one of the few other options)

This means URLs will have `_` in them instead of `%7C`, which I think makes them much more readable.

Death to `%7C`, long live the new flesh

### DIFF
--- a/packages/core/src/nav/common_nav.ts
+++ b/packages/core/src/nav/common_nav.ts
@@ -22,8 +22,10 @@ export const VIEW_SET_HASH = 'viewset';
 export const EMBED_HASH = 'embed';
 /** Prefix for formula pages */
 export const CALC_HASH = 'math';
-/** Path separator */
-export const PATH_SEPARATOR = '|';
+/** The path separator */
+export const PATH_SEPARATOR = '_';
+/** The legacy path separator */
+export const PATH_SEPARATOR_LEGACY = '|';
 /** The query param used to represent the path */
 export const HASH_QUERY_PARAM = 'page';
 
@@ -281,7 +283,7 @@ export function makeUrlSimple(...path: string[]): URL {
 export function makeUrl(navState: NavState): URL {
     const joinedPath = navState.path
         .map(pp => encodeURIComponent(pp))
-        .map(pp => pp.replaceAll(PATH_SEPARATOR, VERTICAL_BAR_REPLACEMENT))
+        .map(pp => pp.replaceAll(PATH_SEPARATOR_LEGACY, VERTICAL_BAR_REPLACEMENT))
         .join(PATH_SEPARATOR);
     const currentLocation = document.location;
     const params = new URLSearchParams(currentLocation.search);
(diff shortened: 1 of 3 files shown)


packages/core/src/nav/common_nav.ts
```diff
@@ -317,16 +319,24 @@ export function splitHashLegacy(input: string) {
...
  *
  * @param input The path, not including ?page=
  */
 export function splitPath(input: string) {
-    return (input.startsWith(PATH_SEPARATOR) ? input.substring(1) : input)
-        .split(PATH_SEPARATOR)
+    let separatorToUse = PATH_SEPARATOR;
+
+    // If this link includes the old separator, use that instead.
+    if (input.includes(VERTICAL_BAR_REPLACEMENT) || input.includes(PATH_SEPARATOR_LEGACY)) {
```
xpdota, line 331: What I'm thinking is that there's no reason to escape pipes if we're using _ as the separator and vice versa, so it should only worry about the first one that it sees.

I think it should still replace _ with a lookalike, since the reason it works without that is specific to the json URLs.

VioletStardustXIV: Is what you're saying that this check should just be `input.includes(PATH_SEPARATOR_LEGACY)`?

xpdota: What I mean is, it should look for the *first* path separator and treat that as the *only* type path separator. i.e. if it encounters a _ first, then it should only look for _. If it encounters a | first, then it should only use |.

VioletStardustXIV:
> Given the large amount of potential consumers ( https://github.com/search?q=%22xivgear.app%22&type=code ) plus several more closed-source ones I'm aware of, I think the best option would be to ninclude the logic _but_ have generated links default to using the existing separator, maybe with an option to enable the new ones for people who want more readable links to be generated.

Honestly, I don't think there's a lot of value in doing it if this is just an option that's not on by default. You can _already_ make the URLs more readable by manually replacing `%7C` but it's really tedious and nobody does it. On the link you posted, if I control + f `%7C` I get 31 results, whereas if I control + f `|` I get three, and none of them are in xivgear links.

The value here is intended to fix the user experience --- users can already edit the links after the fact, and in theory any application that needs to handle xivgear links would need to handle edited links anyway, since they're also valid.

For this to be breaking, the software has to _specifically_ care about parsing generated URLs. Scrolling through, I don't see any of these (including the one with the repo that I own) that will be affected. The API endpoints that are intended to be parsed by computers also won't be affected.

xpdota:
The issue with consumers is that originally, the ability to pass ?page directly to endpoints didn't exist, so they were often manually parsing the URL, e.g. all of the following would fail with the new ones:
* https://github.com/arcan1s/ffxivbis/blob/fa43517b1644293de20e429052f5a5cf25ffe12b/src/main/scala/me/arcanis/ffxivbis/service/bis/parser/impl/XIVGear.scala#L41
* https://github.com/RajahOmen/BisBuddy/blob/5352be62250ee481d687a8ac4c15a05f69255dac/BisBuddy/Services/ImportGearset/XivgearSource.cs#L24
* https://github.com/Yhumi/BisTracker/blob/16a4fd1e3eb34845c1163a6b5c109412060e417e/BisTracker/BiS/BisSheetReader.cs#L39
* https://github.com/XIVLoot/XIVLoot/blob/a5a07d5f60484833d50397607e106d5932401d07/FFXIV-RaidLootAPI/Controllers/PlayerController.cs#L428
* https://github.com/Binboukami/ffxiv-blm-rotation/blob/81b854d01b740366b49d4a53952f320ecd7c9005/src/Components/PlaybackControl.tsx#L626

The reason I suggested having it be optional would just be to get this merged in while consumers are updated over time to accept the new links, while allowing them to test the new link format in their consumers.

VioletStardustXIV:
Hmmm, yeah, that does make sense. For what it's worth, I'm not averse to this specific PR being closed, as it's not a trivial change, but I do think we need to figure Something out for this issue. I do think this has a very severe impact to the UX, and it's why I've used e.g. `https://xivgear.app/#/bis/drk/ultimate/ucob` where I can for readability. 

I much, much, much prefer xivgear to etro in basically every way, but etro's URLs are _way_ way better, and slash-based.

Part of the issue is, creating a v2 URL link while supporting the old v1 would still be breaking, so I do think that fixing this is inevitable to cause some kind of breakage somewhere. For the underscore, it's, at the very least, an easy fix. But I do think that keeping `%7C` as the separator forever would be worse than breaking things at some point.

PR #726, opened 2025-08-14
BLM Sim: using average pps instead of a rotational sim

clo-yunhee:


clo-yunhee:
Ah, i didnt branch right lol

clo-yunhee:
That's not right, I give up for today but the two commits up there are the right ones xD

PR #743, opened 2025-09-02
Add optional buff application delay to actions

VioletStardustXIV:
Adds an optional `buffApplicationDelay` to abilities that, if specified, will take precedence over the ability's `appDelay` for the application of buffs.

### DIFF
--- a/packages/core/src/sims/ability_helpers.ts
+++ b/packages/core/src/sims/ability_helpers.ts
@@ -2,7 +2,9 @@ import {STANDARD_ANIMATION_LOCK, STANDARD_APPLICATION_DELAY} from "@xivgear/xivm
 import {Ability, ComboData} from "./sim_types";
 
 /**
- * Returns the application delay of an ability (from time of snapshot to time of damage/effects applying).
+ * Returns the application delay of an ability (from time of snapshot to time of damage applying).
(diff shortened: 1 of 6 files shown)


packages/core/src/sims/ability_helpers.ts
```diff
@@ -2,14 +2,25 @@ import {STANDARD_ANIMATION_LOCK, STANDARD_APPLICATION_DELAY} from "@xivgear/xivm
...
 
 /**
- * Returns the application delay of an ability (from time of snapshot to time of damage/effects applying).
+ * Returns the application delay of an ability (from time of snapshot to time of damage applying).
+ * By default, if a buffApplicationDelay is not specified, this will be used for the application
+ * delay of the buffs the ability grants, too.
  *
  * @param ability The ability in question
  */
 export function appDelay(ability: Ability) {
```
xiashtra, line 11: Does it make sense to rename this `dmgAppDelay` or something similar now to be more explicit what it's for?

VioletStardustXIV: Oh, that's a good thought, yeah let me fix that

VioletStardustXIV: fixed

PR #820, opened 2025-12-19
[GNB] Revamp of sim for 7.4

VioletStardustXIV:
Works for all level ranges, and doesn't seem to do anything illegal that I can tell.

Without (off by default) unrealistic clip rotation:
image

With unrealistic clip rotation:
image

Keeping this in draft for now while we assess the viability of some 2.49 tech.


xpdota:
Would it make sense to use this as an opportunity to convert it to the new-style gauge system?

PR #835, opened 2025-12-27
[DRK] Fix sim crashing browser tab if total time is specified as 1 or 2

VioletStardustXIV:
We should probably set a minimum of like 5 seconds but I didn't want to make an executive decision while fixing the bug.


xpdota:
Thanks!

PR #872, opened 2026-03-22
Add more Living Shadow/potencyToDamage tests

VioletStardustXIV:
Exports `potencyToDamage` and does some tests with it to validate that values match to expected values.


xpdota:
Thanks!

PR #877, opened 2026-03-30
Level 100 MCH simulator

zodiia:
This simulator should be working for any GCD at level 100. I plan on adding support for lower levels as well, but we're not there yet lol. 

I have yet to implement the automaton queen scaling, I haven't been able to figure this one out, will need help with that. Everything should otherwise be good!


packages/frontend/src/scripts/sims/ranged/mch_sim_gui.ts
```diff
@@ -0,0 +1,123 @@
+import {FieldBoundCheckBox, labeledCheckbox} from "@xivgear/common-ui/components/util";
+import {BaseMultiCycleSimGui} from "../multicyclesim_ui";
+import {AbilitiesUsedTable} from "../components/ability_used_table";
+import {CycleSimResult, DisplayRecordFinalized, isFinalizedAbilityUse} from "@xivgear/sims/cycle_sim";
+import {ColDefs, CustomColumnSpec} from "@xivgear/common-ui/table/tables";
+import {MchSimSettings, MchSimResult} from "@xivgear/sims/ranged/mch/mch_sheet_sim";
+import type {MchPreDmgUsedAbility} from "@xivgear/sims/ranged/mch/mch_types";
+
+// Fully copied from https://github.com/xiv-gear-planner/gear-planner/pull/556
```
VioletStardustXIV: I think let's delete this line, it's likely to get forgotten about

packages/sims/src/common/cooldown_manager.ts
```diff
@@ -134,7 +134,12 @@ export class CooldownTracker {
         }
         const key = cooldownKey(ability);
         const cdState = this.currentState.get(key);
-        this.currentState.set(cooldownKey(ability), new InternalState(cdState.cappedAt + delta));
+        if (cdState) {
```
VioletStardustXIV, line 137: Pure interest reasons: why would cdState be falsey here?

packages/sims/src/ranged/mch/mch_actions.ts
```diff
@@ -0,0 +1,253 @@
...
+    appDelay: 0,
+    id: 16504,
+    potency: 2.4, // multiply by battery used
+    attackType: 'Ability',
+} as const;
+
+export const AutomatonQueenPileBunker: MchOgcdAbility = {
+    type: 'ogcd',
+    name: '(Automaton Queen) Pile Bunker',
+    alternativeScalings: ['Pet Action Weapon Damage'],
```
VioletStardustXIV: Should these have `'Automaton Queen Dexterity Scaling'` too? (I know it's not implemented yet)

packages/sims/src/ranged/mch/mch_actions.ts
```diff
@@ -0,0 +1,253 @@
...
+    attackType: 'Ability',
+    activatesBuffs: [OverheatedBuff],
+    updateGauge: (gauge) => gauge.heat -= 50,
+    cooldown: {
+        time: 10,
+    },
+} as const;
+
+// Check how to apply wildfire damage
+// also remember that wildfire cannot dh/crit
```
VioletStardustXIV: One option here would be to use BuffEffects, and something like
```
critChanceIncrease: -1
dhitChanceIncrease: -1
```

packages/sims/src/ranged/mch/mch_buffs.ts
```diff
@@ -0,0 +1,71 @@
...
+    selfOnly: true,
+    appliesTo: (ability) => ability.name === 'Excavator',
+    beforeSnapshot: (buff) => buff.removeSelf(),
+};
+
+export const WildfireBuff: Buff = {
+    name: 'Wildfire',
+    duration: 10,
+    statusId: 1946,
+    effects: {},
```
VioletStardustXIV, line 68: Maybe `critChanceIncrease: -1` and `dhitChanceIncrease: -1`?

packages/sims/src/ranged/mch/mch_sheet_sim.ts
```diff
@@ -0,0 +1,626 @@
...
+    // if set to true, cooldowns like air anchor won't be held to align on 2-min bursts
+    dontAlignCds: boolean;
+}
+
+export interface MchSimSettingsExternal extends ExternalCycleSettings<MchSimSettings> {}
+export interface MchSimResult extends CycleSimResult {}
+
+export const mchSheetSpec: SimSpec<MchSheetSim, MchSimSettingsExternal> = {
+    stub: 'mch-sheet-sim',
+    displayName: 'MCH simulator',
```
VioletStardustXIV:
```suggestion
    displayName: 'MCH Sim',
```

VioletStardustXIV: Consistency :)

packages/sims/src/ranged/mch/mch_sheet_sim.ts
```diff
@@ -0,0 +1,626 @@
...
+        }
+        return false;
+    }
+
+    // Always use
+    private canUseDrill(): boolean {
+        if (this.cdTracker.statusOf(Drill).currentCharges === 2) {
+            console.warn(`[${formatDuration(this.currentTime)}] Drill overcapped`);
+        }
+        if (!this.simSettings.dontAlignCds && this.cdTracker.statusOf(Drill).cappedAt.relative > this.timeBeforeNextBurstWindow(this.gcdCalculated, true)) { // keep a charge for burst
```
VioletStardustXIV: nitpick: move the comments to above the conditions

alternatively: replace the conditions with a helper method, e.g.
if(!keepAChargeOfDrillForBurst(parameters)

packages/sims/src/ranged/mch/mch_buffs.ts
VioletStardustXIV: wait no that wouldn't work lol that'd affect all GCDs not just Wildfire

VioletStardustXIV: Could use a second thing that only applies to `Detonator` perhaps?

packages/sims/src/common/cooldown_manager.ts
zodiia: It seems it might not be useful anymore? At first, when using the cooldown tracker for the pot (by tricking it into thinking it has a 5/6 minutes cooldown) it seemed that in this spot specifically it could be falsey for the pot. After testing again now, it looks like it doesn't do that anymore? I'll keep that condition just in case (and also because it doesn't break existing logic), but if you want me to remove it if it's not needed anymore, I can do that as well.
If not I'll let you resolve this yourself!

xpdota: cdState would be falsey if the cooldown hasn't been used normally yet.

xpdota: However, I will add that this behavior should be clearly documented, as "[modify] the cooldown of the given ability by the ddelta specified" doesn't really make sense if the cooldown isn't rolling at all.

zodiia: Right, I use it to "modify" the cooldown of the pot by 2 minutes for the "start pots at 2 minutes" option. Without that check, and since I haven't used the potion before, it will result in an error. 

I could rewrite my pot logic to not use the cd tracker at all, but using it was simpler, at least for me.
For now I'll add documentation for this, unless you want me to remove that behavior completely.

packages/sims/src/ranged/mch/mch_actions.ts
xpdota: I think it would make more sense to do -999 or something like that so that a crit chance buff would never be able to push it back above 0.

VioletStardustXIV (approved):
Looks great, thanks! :)

packages/sims/src/common/cooldown_manager.ts
xpdota: I would say it makes more sense to do the logic within the cycle processor rather than the cooldown manager. For example, if we later add the option to show cooldown states in the results display, it wouldn't make sense to say that the potion is on cooldown.

PR #889, opened 2026-04-28
[MCH] Automaton queen scaling + SkS multi for Wildfire

VioletStardustXIV:
Adds correct automaton queen scaling, and properly scales Wildfire with Skill Speed

### DIFF
--- a/packages/core/src/sims/sim_types.ts
+++ b/packages/core/src/sims/sim_types.ts
@@ -296,7 +296,10 @@ export type ComboBehavior = ComboData['comboBehavior'];
  * Alternate scalings that can exist for abilities, e.g. Living
  * Shadow, Bunshin, SMN pet actions.
  */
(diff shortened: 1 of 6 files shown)


packages/core/src/sims/sim_types.ts
```diff
@@ -296,7 +296,10 @@ export type ComboBehavior = ComboData['comboBehavior'];
  * Alternate scalings that can exist for abilities, e.g. Living
  * Shadow, Bunshin, SMN pet actions.
  */
-export type AlternativeScaling = "Living Shadow Strength Scaling" | "Automaton Queen Dexterity Scaling" | "Pet Action Weapon Damage";
+export type AlternativeScaling = "Living Shadow Strength Scaling" |
```
xpdota, line 299: I don't know if there's a lint for this (would be nice if it exists), but I prefer the operator be on the start of a following line, i.e.
```
thing = one
    | two
    | three
```

PR #900, opened 2026-05-11
change rpr sim default time to 510s

Junie-774:
changes rpr sim's default time to 510s (8:30). RPR's standard rotation only works up to but not including 10:30, and longer full-uptime KT's are rare, so 8:30 is a better default than 12:00


xpdota:
Thanks!

PR #901, opened 2026-05-11
override default cycle settings instead

Junie-774:
Better way to handle the previous PR changing RPR's default sim time


xpdota:
Thanks!

PR #924, opened 2026-07-27
Extend AST support to SB and ShB ults

jtcasper:
Compared with previous 100 sim, no changes. Previous 90 sim it turns out had been using 6.0 Combust III potency (55/tick) which has been 65/tick at 90 since 7.0.

### DIFF
--- a/packages/sims/src/healer/ast_sheet_sim.ts
+++ b/packages/sims/src/healer/ast_sheet_sim.ts
@@ -40,41 +40,69 @@ export type AstExtraData = {
 }
 
 const filler: AstGcdAbility = {
-    id: 25871,
+    id: 7442,
     type: 'gcd',
-    name: "Fall Malefic",
-    potency: 250,
+    name: "Malefic III",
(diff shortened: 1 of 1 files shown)


packages/sims/src/healer/ast_sheet_sim.ts
```diff
@@ -40,41 +40,69 @@ export type AstExtraData = {
...
+        },
+        {
+            minLevel: 82,
+            id: 25871,
+            name: "Fall Malefic",
+            potency: 250,
+        },
         {
             minLevel: 94,
+            id: 25871,
```
xpdota, line 65: I believe you can drop these since the second levelModifier has no max level, so only potency should be required.

jtcasper: Last time I did this I remembered the `LevelModifier`s not being applied in a "stacked manner", but the "highest applicable one" overwriting the base ability. Did that get changed?

